PR #1, opened 2020-08-18
Port release_90 patches to LLVM 10 monorepo

bryanpkc:
This pull request cherry-picks all relevant patches from the `release_90` branches in https://github.com/flang-compiler/flang-driver and https://github.com/flang-compiler/llvm, adapts them to fit the monorepo layout, and adjusts the Classic Flang code to let it coexist with the new Flang. To build the code, run CMake with `-DLLVM_ENABLE_CLASSIC_FLANG=on`.

See https://github.com/Huawei-PTLab/classic-flang-llvm/wiki for an example build script.

PR #4, opened 2020-10-12
[ClassicFlang] [Driver] Reduce downstream delta

peterwaller-arm:
This commit is motivated by reducing the merge burden by shrinking the diff
between llvm upstream and classic-flang-llvm-project.

Outside of Flang, fortran code is fed through the Compile phase, and the
appropriate tooling is picked up through ToolChain::SelectTool.

Classic flang introduced a FortranFrontend, but these days this seems
unnecessary. Fortran can go through the same machinary as everything else.

The main thing classic flang does differently is to have the compiler phase do
the preprocessing.

* Use the Compile phase to compile fortran code.
* Remove FortranFrontendJobClass.
* Remove FortranFrontend tool (instead it's just the Flang tool, which in
  classic flang mode is classic flang).
* When compiling with USE_CLASSIC_FLANG, assume IsFlangMode == true.
  * This makes the driver go through the "new" fortran code paths rather than
    the old gfortran/gnu fallback in ToolChain::SelectTool.
* Update a test which inspects the output of the flang tooling.
  * Previously the Compile stage emitted IR. In the interests of avoiding an
    increase of the delta from upstream llvm-project, follow what they do and
    pass bitcode downwards.

# Status

This PR is draft and I've run out of time for it, so it's up for adoption.. It doesn't yet give a correct result (`make check-clang-flang-driver`), I believe because the driver thinks flang is capable of emitting assembly, and therefore combines the stages incorrectly. Please see [my comments here](https://github.com/flang-compiler/classic-flang-llvm-project/pull/1#issuecomment-705598317).

### DIFF
--- a/clang/include/clang/Driver/Action.h
+++ b/clang/include/clang/Driver/Action.h
@@ -62,7 +62,6 @@ public:
     AnalyzeJobClass,
     MigrateJobClass,
     CompileJobClass,
-    FortranFrontendJobClass,
     BackendJobClass,
     AssembleJobClass,
     LinkJobClass,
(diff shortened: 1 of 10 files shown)


bryanpkc:
Oops, I didn't see this PR before creating #5. I am going to close this one as the new PR addresses a number of issues, including the one you pointed out in https://github.com/flang-compiler/classic-flang-llvm-project/pull/1#issuecomment-705598317.

peterwaller-arm:
Thank you @bryanpkc, very much appreciated :)

PR #5, opened 2020-10-12
[ClassicFlang][Driver] Reduce downstream delta

bryanpkc:
This commit is motivated by reducing the merge burden by shrinking the
diff between llvm upstream and classic-flang-llvm-project.

Outside of Flang, Fortran code is fed through the Compile phase, and
the appropriate tooling is picked up through ToolChain::SelectTool.

Classic Flang introduced a FortranFrontend, but these days this seems
unnecessary. Fortran can go through the same machinery as everything else.

* Use the Preprocess phase to preprocess Fortran code. This phase is
  always combined with the Compile phase.

* Use the Compile phase to lower Fortran code to LLVM IR, and use the
  Backend phase to compile and optimize the IR. These phases are never
  combined.

* Remove FortranFrontendJobClass.

* Remove FortranFrontend tool (instead it's just the Flang tool, which
  in Classic Flang mode is Classic Flang).

* Update tests which inspect the output of the Classic Flang tooling,
  and ensures that the driver does the right thing for various types of
  inputs.

Based on a patch by @peterwaller-arm (peterwaller-arm@10dacac072).


bryanpkc:
@banach-space Thanks for volunteering to review.

clang/test/Driver/flang/classic_flang.f95
```diff
@@ -1,28 +1,44 @@
...
+
+! RUN: %clang --driver-mode=flang -target x86_64-unknown-linux-gnu -E %s -### 2>&1 \
+! RUN:   | FileCheck --check-prefix=CHECK-PREPROCESS %s
+! CHECK-PREPROCESS: flang1
+! CHECK-PREPROCESS-SAME: "-preprocess"
+! CHECK-PREPROCESS-SAME: "-es"
+! CHECK-PREPROCESS-SAME: "-pp"
+! CHECK-PREPROCESS-NOT: flang1
+! CHECK-PREPROCESS-NOT: flang2
+! CHECK-PREPROCESS-NOT: clang
```
peterwaller-arm: I recommend - where you can - enclosing all of these (and elsewhere) in quotes. We've seen an issue in the past where tests fail because the working directory contains, e.g. `clang`, because maybe the directory name is based on a branch name, or a user name, and suddenly the test fails.

clang/lib/Driver/Driver.cpp
```diff
@@ -4041,6 +4017,10 @@ class ToolSelector final {
     if (!T)
       return nullptr;
 
+    // Classic Flang is not integrated with the backend.
+    if (C.getDriver().IsFlangMode() && !T->hasIntegratedAssembler())
+      return nullptr;
+
```
peterwaller-arm, line 4023: Need an `#ifdef ENABLE_CLASSIC_FLANG` ?

clang/lib/Driver/Driver.cpp
```diff
@@ -3306,10 +3282,13 @@ void Driver::handleArguments(Compilation &C, DerivedArgList &Args,
       if (InputArg->isClaimed())
         continue;
 
-      // Fortran input is preprocessed using the frontend.
-      if (InitialPhase == phases::FortranFrontend &&
+#ifdef ENABLE_CLASSIC_FLANG
+      // If the input is detected as already preprocessed (e.g. has the .f95
+      // extension), and the user specifies -E, preprocess the file anyway.
+      if (IsFlangMode() && InitialPhase == phases::Compile &&
```
banach-space, line 3288: I'm not a Fortran expert, but `gfortran` docs suggest that 
* `.f95` means that the file _has not been preprocessed yet_

See here: https://gcc.gnu.org/onlinedocs/gcc/Overall-Options.html

Also, IIUC, this block of code makes sure that the parsing of preprocessor specific options is not happening here, i.e. in the compiler driver. Instead, it should happen in the frontend driver, i.e. where the preprocessor is run. So it's not about whether the preprocessor is run or not, but about whether to further parse the options or not.

I might be wrong though!

bryanpkc: @banach-space In the page you linked, the description for `.f95` says:

> <dl>
> <dt>*file*.f95</dt>
> <dd>Free form Fortran source code that should not be preprocessed.</dd>
> </dl>

The description is consistent with original Classic Flang code, as well as this patch. The uppercase `.F95` indicates that a file must be preprocessed, and the output of the preprocessor is typically named with `.f95`.

This block of code is a hack to maintain backward compatibility with legacy behaviour, i.e. allow `-E` to force preprocessing on files that are normally not preprocessed. If we can reach a consensus about not supporting this behaviour, I would prefer to get rid of this hack, and let the driver print a "previously preprocessed input" warning and do nothing. However, that is a bigger decision and should not be done in this PR.

clang/test/Driver/flang/classic_flang.f95
bryanpkc: OK I will turn these into regular expressions.

clang/lib/Driver/Driver.cpp
bryanpkc: `T->hasIntegratedAssembler` returns `true` for the new Flang, so this works without a macro, which I thought was preferable. It would be great if this could be upstreamed into llvm-project.

clang/lib/Driver/Driver.cpp
banach-space: I misread that, sorry! This clarifies a lot @bryanpkc, thank you !

This makes sense - we should preserve the current behavior.

banach-space (commented):
[nit] @bryanpkc On systems where filenames are case-insensitive (e.g. OS X), `classic_flang.F95`  and `classic_flang.f95` will be problematic (it's treated as one file).  Just to be on the safe side, perhaps it's worth to rename them?

bryanpkc:
> Just to be on the safe side, perhaps it's worth to rename them?

@banach-space The use of uppercase file name extensions to indicate the need for preprocessing seems to be a common policy observed by many Fortran compiler vendors, including [GCC](https://gcc.gnu.org/onlinedocs/gcc/Overall-Options.html), [PGI](http://geco.mines.edu/guide/pgf90.html), [Cray](https://pubs.cray.com/bundle/Cray_Fortran_Reference_Manual_85_S-3901_Fortran_ditaval.xml/page/Source_Preprocessing.html), and [Intel](https://software.intel.com/content/www/us/en/develop/documentation/fortran-compiler-developer-guide-and-reference/top/compiler-setup/using-the-command-line/understanding-file-extensions.html). I think we should open an issue to discuss how to accommodate case-insensitive file systems, but that work wouldn't belong in this PR.

banach-space:
> @banach-space The use of uppercase file name extensions to indicate the need for preprocessing seems to be a common policy observed by many Fortran compiler vendors, including [GCC](https://gcc.gnu.org/onlinedocs/gcc/Overall-Options.html), [PGI](http://geco.mines.edu/guide/pgf90.html), [Cray](https://pubs.cray.com/bundle/Cray_Fortran_Reference_Manual_85_S-3901_Fortran_ditaval.xml/page/Source_Preprocessing.html), and [Intel](https://software.intel.com/content/www/us/en/develop/documentation/fortran-compiler-developer-guide-and-reference/top/compiler-setup/using-the-command-line/understanding-file-extensions.html). I think we should open an issue to discuss how to accommodate case-insensitive file systems, but that work wouldn't belong in this PR.

Sorry, I should've been clearer :) I'm not suggesting any extra work nor discussion. I agree that that would be beyond the scope of this patch. I'm definitely not suggesting that we challenge Fortran's file-naming conventions.

I was suggesting that the test files are renamed (which, AFAIK, doesn't violate any of the file-naming conventions):
* `classic_flang.F95` -> `classic_flang_not_preprocessed.F95`
* `classic_flang.f95` -> `classic_flang_preprocessed.f95`
* `classic_flang.F` -> `classic_flang_not_preprocessed.F`
* `classic_flang.f` -> `classic_flang_preprocessed.f`

Otherwise, your patch/commit confuses the filesystem on OS X. However, AFAIK, OS X *is not* officially supported by Flang, so for me this renaming would be a _nice to have_. 

This patch LGTM as is, thanks for working on this!

gklimowicz:
> I was suggesting that the test files are renamed (which, AFAIK, doesn't violate any of the file-naming conventions):
> 
> * `classic_flang.F95` -> `classic_flang_not_preprocessed.F95`
> * `classic_flang.f95` -> `classic_flang_preprocessed.f95`
> * `classic_flang.F` -> `classic_flang_not_preprocessed.F`
> * `classic_flang.f` -> `classic_flang_preprocessed.f`
> 

Unless I'm letting myself get confused by your request, I think it's the other way around: `.F` and `.F95` invoke preprocessing. `.f` and `.f95` do not invoke the preprocessor.

So wouldn't it be this?

* `classic_flang.F95` -> `classic_flang_preprocessed.F95`
* `classic_flang.f95` -> `classic_flang_not_preprocessed.f95`
* `classic_flang.F` -> `classic_flang_preprocessed.F`
* `classic_flang.f` -> `classic_flang_not_preprocessed.f`

banach-space:
> Unless I'm letting myself get confused by your request, I think it's the other way around: `.F` and `.F95` invoke preprocessing. `.f` and `.f95` do not invoke the preprocessor.
> 
> So wouldn't it be this?
> 
>     * `classic_flang.F95` -> `classic_flang_preprocessed.F95`
> 
>     * `classic_flang.f95` -> `classic_flang_not_preprocessed.f95`
> 
>     * `classic_flang.F` -> `classic_flang_preprocessed.F`
> 
>     * `classic_flang.f` -> `classic_flang_not_preprocessed.f`

Good point - does `classic_flang_preprocessed.F95` mean that the file:
*  _has been preprocessed_ or,
*  _should be preprocessed_?
The naming that I suggested is ambiguous, sorry! 

Perhaps `class_flang_do_preprocess.F95` and `classic_flang_do_not_preprocess.f95` instead? Does that make more sense?

gklimowicz:
> Good point - does `classic_flang_preprocessed.F95` mean that the file:
> 
> * _has been preprocessed_ or,
> * _should be preprocessed_?
>   The naming that I suggested is ambiguous, sorry!

_should be_ preproprocessed.

> Perhaps `class_flang_do_preprocess.F95` and `classic_flang_do_not_preprocess.f95` instead? Does that make more sense?

It certainly is the right intent. I just hope I don't have to type them! :-)

bryanpkc:
I have named the test files differently. I decided to let `classic-flang.f95` remain short and simple, as I expect more driver-related tests to be added there.

clang/test/Driver/flang/classic_flang.f95
bryanpkc: Fixed.

bryanpkc:
Pinging reviewers @banach-space @peterwaller-arm @kiranchandramohan @gklimowicz @shivaramaarao

banach-space (approved):
Thank you for working on this @bryanpkc ! LGTM. From what I can tell, the CI failures are unrelated to the changes in this patch.

gklimowicz (approved):
This builds and passes our tostones on OpenPower and LLVM 10.

bryanpkc:
@kiranchandramohan @shivaramaarao  Could you review this PR?

kiranchandramohan (approved):
LGTM.

bryanpkc:
@shivaramaarao Could you review this PR please?

shivaramaarao (approved):
looks good to me. thanks.

PR #8, opened 2020-11-23
[DebugInfo] Backporting multiple upstream commits supporting fortran array debuggability

alokkrsharma:
This pull request is for back-porting already committed patches to upstream LLVM.

https://reviews.llvm.org/D79306 ([DebugInfo] llvm rejects DWARF operator DW_OP_push_object_address)
https://reviews.llvm.org/D79592 ([DebugInfo] support for DW_AT_data_location in llvm)
https://reviews.llvm.org/D80197 ([DebugInfo] Upgrade DISubrange to support Fortran dynamic arrays)
https://reviews.llvm.org/D83544 ([DebugInfo] Support for DW_AT_associated and DW_AT_allocated.)
https://reviews.llvm.org/D87500 ([DebugInfo] DISubrange support for fortran assumed size array)
https://reviews.llvm.org/D89141 [DebugInfo] Support for DWARF attribute DW_AT_rank
https://reviews.llvm.org/D89208 [DebugInfo] Support for DWARF operator DW_OP_over
https://reviews.llvm.org/D89218 [DebugInfo] Support for DW_TAG_generic_subrange

PR #9, opened 2020-11-30
[ClassicFlang] Fix DIFortranSubrange.getLowerBound assert failure

bryanpkc:
An `ALLOCATABLE` array that also has the `SAVE` attribute has a section descriptor that resides in the BSS, and its debuginfo is encoded as a `DIGlobalVariableExpression`. However, the `DIVariable`-dependent `DIFortranSubrange` metadata fails an assertion if the descriptor's debuginfo metadata is not a `DIVariable`.

The following example triggers the problem (`POINTER, SAVE` arrays are also similarly affected):
```fortran
subroutine test(i)
  integer :: i
  real, dimension(:), allocatable, save :: arr
  allocate(arr(i))
end subroutine
```

### DIFF
--- a/llvm/include/llvm/IR/DebugInfoMetadata.h
+++ b/llvm/include/llvm/IR/DebugInfoMetadata.h
@@ -382,15 +382,11 @@ public:
 
   TempDIFortranSubrange clone() const { return cloneImpl(); }
 
-  DIVariable *getLowerBound() const {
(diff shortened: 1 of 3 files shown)


gklimowicz (approved):
This passes our tests on OpenPOWER for LLVM 10.

bryanpkc:
This is no longer needed in light of #8 and flang-compiler/flang#952.

PR #12, opened 2021-01-24
Github Actions to build release_11x

michalpasztamobica:
Add `release_11x` to CI scripts. 
`release_100` is still build the same way. 
Artifacts have branch name suffix for `flang` CI to be able to distinguish them.

### DIFF
--- a/.github/workflows/pre-compile_llvm.yml
+++ b/.github/workflows/pre-compile_llvm.yml
@@ -2,7 +2,7 @@ name: Pre-compile llvm
 
 on:
   push:
-    branches: [ release_100 ]
+    branches: [ release_100, release_11x ]
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


michalpasztamobica:
@kiranchandramohan , @bryanpkc , @RichBarton-Arm , @shivaramaarao, @mleair, please review or help me set up reviewers appropriately. Thanks.

.github/workflows/pre-compile_llvm.yml
```diff
@@ -50,14 +50,17 @@ jobs:
           make --version
           ${{ matrix.cc }}-${{ matrix.version }} --version
           ${{ matrix.cpp }}-${{ matrix.version }} --version
-      
+      - name: Extract branch name
+        run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
```
bryanpkc: Just curious, [GitHub documentation](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idoutputs) seems to suggest using the `::set-output name=branch::` syntax in this situation. Can you point me to the documentation that discusses the `##[set-output name=branch;]` syntax?

Also, the final `$(echo ${GITHUB_REF#res/heads/})` seems redundant. Couldn't the shell interpolate `${GITHUB_REF#refs/heads/}` within the first `echo` directly?

michalpasztamobica: Good point, @bryanpkc . The command works with one `echo` less and I changed to the yaml syntax (`::`) as suggested in the docs. 
I found the other approach on [stackoverflow](https://stackoverflow.com/questions/58033366/how-to-get-current-branch-within-github-actions), but you're right that it is not documented anywhere, so perhaps it is some outdated, but still working API. Github Actions are changing quite fast...
Thanks for the feedback!

bryanpkc (approved):
LGTM, thanks @michalpasztamobica.

michalpasztamobica:
@kiranchandramohan , @shivaramaarao, could you let me know your toughts on this and the "twin" PR#13?

shivaramaarao (approved):
LGTM. thanks

kiranchandramohan (approved):
LGTM.

PR #15, opened 2021-02-11
Fortran dynamic array rel11x

michalpasztamobica:
Test CI


michalpasztamobica:
This started CI, so the fix should be valid.

PR #21, opened 2021-02-17
CCache LLVM precompilation on GitHub Actions

janisozaur:


janisozaur:
As the GH Action is only ran on `release_*` branches, I've started the test in my own fork: https://github.com/janisozaur/classic-flang-llvm-project/actions/runs/575978966, will report back with results once finished.

janisozaur:
There's something not quite up to snuff with this setup, still investigating

bryanpkc:
We are using ccache in our CI workflows, so I guess this is no longer needed.

PR #27, opened 2021-03-17
Allow full path to flang1 and flang2 in classic flang tests.

michalpasztamobica:
When testing @bryanpkc 's release_12x branch I noticed the four tests were failing for me because the call to flang1 and flang2 contained a full path of the tool, e.g. `/usr/bin/local/flang1`. Somehow this failed to match:

```
! CHECK: "flang1"
```

I tried building with different cmake, make, ninja, gcc and clang version and tried building completely outside of the source tree, just outside llvm or using the main cmake script as indicated in the llvm-project instructions - always with the same result.

I do not see any reason why this should be considered an invalid call so I propose we add a wildcard to the beginning of the expected string to indicate that this is OK.

### DIFF
--- a/clang/test/Driver/flang/classic-flang.f
+++ b/clang/test/Driver/flang/classic-flang.f
@@ -5,10 +5,10 @@
 
 ! RUN: %clang --driver-mode=flang -target x86_64-unknown-linux-gnu -c %s -### 2>&1 \
 ! RUN:   | FileCheck %s
-! CHECK: "flang1"
+! CHECK: "{{.*}}flang1"
 ! CHECK-NOT: "-preprocess"
 ! CHECK-SAME: "-nofreeform"
-! CHECK-NEXT: "flang2"
+! CHECK-NEXT: "{{.*}}flang2"
 ! CHECK-NEXT: {{clang.* "-cc1"}}
 
 ! Check that the driver invokes flang1 correctly when preprocessing is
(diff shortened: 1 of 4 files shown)


clang/test/Driver/flang/classic-flang.f
```diff
@@ -5,22 +5,22 @@
...
 ! RUN:   | FileCheck --check-prefix=CHECK-PREPROCESS %s
-! CHECK-PREPROCESS: "flang1"
+! CHECK-PREPROCESS: "{{.*}}flang1"
 ! CHECK-PREPROCESS-SAME: "-preprocess"
 ! CHECK-PREPROCESS-SAME: "-es"
 ! CHECK-PREPROCESS-SAME: "-pp"
-! CHECK-PREPROCESS-NOT: "flang1"
-! CHECK-PREPROCESS-NOT: "flang2"
+! CHECK-PREPROCESS-NOT: "{{^.*}}flang1"
+! CHECK-PREPROCESS-NOT: "{{^.*}}flang2"
```
bryanpkc, line 24: Question: Why is the `^` needed in the regexps here? I removed the caret and the tests still work.

bryanpkc (approved):
LGTM. Thanks for fixing my bug (I introduced the tests and even worked around this very problem in the `"clang"` case, but didn't realize it would also affect "flang1" and "flang2").

pawosm-arm (approved):
Looks OK. It's funny how we could do that without this change for such a long time...

SouraVX (approved):

alokkrsharma (approved):
Looks good, thanks.

clang/test/Driver/flang/classic-flang.f
michalpasztamobica: Indeed, the caret should not be here. I added it in the beginning, but when I got how it works I removed it from all expressions. I missed this one usage. 
I don't want to set up a new PR now, just for this one change. Can we agree that whoever changes the tests in the nearest future removes this caret? :)

pawosm-arm: sadly, this is a recipe for leaving it like that forever.

bryanpkc: @pawosm-arm It would have been nice to not accept pull requests that have open, unresolved discussions.

bryanpkc: But then maybe I shouldn't have approved it preemptively.

michalpasztamobica: It is trivial to fix the errors, so I just set up two more PRs:
https://github.com/flang-compiler/classic-flang-llvm-project/pull/29
https://github.com/flang-compiler/classic-flang-llvm-project/pull/30
Apologies for not noticing this earlier and thus - polluting the commit history.
@bryanpkc , in case you cherry-pick this to `release_12x`, please feel free to squash the extra commit with the original one.

PR #28, opened 2021-03-17
Allow full path to flang1 and flang2 in classic flang tests (release_11x)

michalpasztamobica:
This is essentially the same as https://github.com/flang-compiler/classic-flang-llvm-project/pull/27, just set for `release_11x` branch.

When testing @bryanpkc 's `release_12x` branch I noticed the four tests were failing for me because the call to flang1 and flang2 contained a full path of the tool, e.g. `/usr/bin/local/flang1`. Somehow this failed to match:

```
! CHECK: "flang1"
```

I tried building with different cmake, make, ninja, gcc and clang version and tried building completely outside of the source tree, just outside llvm or using the main cmake script as indicated in the llvm-project instructions - always with the same result.

I do not see any reason why this should be considered an invalid call so I propose we add a wildcard to the beginning of the expected string to indicate that this is OK.

### DIFF
--- a/clang/test/Driver/flang/classic-flang-must-preprocess.F
+++ b/clang/test/Driver/flang/classic-flang-must-preprocess.F
@@ -5,8 +5,8 @@
 
 ! RUN: %clang --driver-mode=flang -target x86_64-unknown-linux-gnu -c %s -### 2>&1 \
 ! RUN:   | FileCheck %s
-! CHECK: "flang1"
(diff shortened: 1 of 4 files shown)


pawosm-arm (approved):
Assuming this is the same thing on a different branch.

SouraVX (approved):
LGTM, Thanks!

PR #31, opened 2021-03-21
Testing GitHub workflow on release_12x

bryanpkc:
Do not accept this PR.


michalpasztamobica:
Hi @bryanpkc . I think you need to change something in `/llvm` and `/clang` to trigger the PR workflows. See [here why](https://github.com/flang-compiler/classic-flang-llvm-project/blob/release_100/.github/workflows/llvm-tests.yml#L10-L16).

PR #33, opened 2021-04-19
Remove the builds for gcc-9

michalpasztamobica:
This is a consequence of https://github.com/flang-compiler/flang/pull/1012. We don't need the gcc-9 builds any more.
I intentionally left the checks for gcc version, even though we now only use gcc-10, in case we want to add gcc-11 in the future.

### DIFF
--- a/.github/workflows/pre-compile_llvm.yml
+++ b/.github/workflows/pre-compile_llvm.yml
@@ -12,12 +12,8 @@ jobs:
         target: [X86]
         cc: [clang]
         cpp: [clang++]
-        version: [9, 10, 11]
+        version: [10, 11]
         include:
-          - target: X86
-            cc: gcc
-            cpp: g++
-            version: 9
           - target: X86
             cc: gcc
             cpp: g++


kiranchandramohan:
Do we need clang-9?
```
        cpp: [clang++]
          version: [9, 10, 11]
```

Do we need the update-alternatives for gcc-9 ?
```
      - if: matrix.cc == 'gcc' && matrix.version == '10'
        run: |
          sudo add-apt-repository ppa:ubuntu-toolchain-r/ppa
          sudo apt install gcc-10 g++-10
          sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-9 10
          sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-10 20
          sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9 10
```

michalpasztamobica:
We still need the `update-alternatives` because gcc-9 comes as a default in ubuntu-20.04. So we need to set gcc-10 as the first choice after installing it.

For the llvm-9 - we wanted it half a year ago, but I don't know if we still need it. Perhaps we can put this on hold and ask in the Wednesday call?

kiranchandramohan (approved):
OK. LGTM.

bryanpkc (approved):
LGTM.

michalpasztamobica:
I just force-pushed a change that removes the llvm-9 compiler from the build matrix, as discussed in the call.

Please, do not merge this yet, I need to provide a related PR to flang, so we don't break the CI there (it checks if the classic-flang-llvm-project built 5 artifacts per release, but it will now only build 3 (llvm-10, llvm-11 and gcc-10), so we need to merge the two closely one after another in order not to break anything.

michalpasztamobica:
@kiranchandramohan , I just submitted https://github.com/flang-compiler/flang/pull/1035 - it should build fine after we merge this PR. We have a safety buffer of 10 commits on each branch until flang starts failing due to insufficient number of artifacts (we will now build 3, instead of 5). The abovementioned PR fixes this.

PR #36, opened 2021-05-05
Build and test Flang project on PRs

michalpasztamobica:
As discussed in the Bi-weekly Classic Flang sync - it's a good idea to make sure that changes to `classic-flang-llvm-project` do not break the `flang` CI. This job will run in parallel to the clang and llvm tests and should not take up more time as it is using ccache to speed up the build.
I had some issues with ccache, most likely to a cancelled run that keeps a lock on some key, I hope the issue will be gone in this repo, but in the worst case I will change this to draft.

We are considering more changes to how CI is running, for example moving the commands to bash scripts to allow easy local reproduction of failing cases, but I prefer to introduce changes step by step, to get meaningful reviews.

### DIFF
--- a/.github/workflows/flang-tests.yml
+++ b/.github/workflows/flang-tests.yml
@@ -1,0 +1,126 @@
+name: Build and test Flang
+
+on:
+  pull_request:
+    branches: [ release_100, release_11x, release_12x ]
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    env:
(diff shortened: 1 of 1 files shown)


pawosm-arm (approved):

SouraVX (approved):

michalpasztamobica:
Hi @bryanpkc , does this PR look OK to you as well? :)

bryanpkc (approved):

PR #41, opened 2021-05-10
Prepare llvm and clang for ARM64 build of Flang

michalpasztamobica:
Run Dockerfile script on a self-hosted ARM64 runner to prepare a docker image with llvm and clang built and installed for Flang to use when testing code.

PR #44, opened 2021-05-20
Add missing <limits> header

michalpasztamobica:
While working on another Pull-request I noticed the CI is failing. This coincides with Ubuntu-20.04.2 update. I did not manage to reproduce the failure on a local docker repository, even after setting up the exact same tools, that github mentions to install for their docker image.

The root cause of the failure is that one of the header files is missing the #include clause. Somehow the tools which we are normally using managed to go around it and build fine, but the current github's Ubuntu image has a set of tools which is not able to cover this and revealed the flaw.

I think the best solution is to add the missing header and not count on the build tools to do this for us.

I submit a separate PR in order not to mix it with the CI changes I am trying to commit in the original PR.

Essentially identical to #42, just for another branch.

### DIFF
--- a/llvm/utils/benchmark/src/benchmark_register.h
+++ b/llvm/utils/benchmark/src/benchmark_register.h
@@ -1,6 +1,7 @@
 #ifndef BENCHMARK_REGISTER_H
 #define BENCHMARK_REGISTER_H
 
+#include <limits>
 #include <vector>
 
 #include "check.h"


pawosm-arm (approved):

michalpasztamobica:
@bryanpkc , @SouraVX , I'd be very grateful if you had a few seconds to review this tiny change and if possible also look into https://github.com/flang-compiler/classic-flang-llvm-project/pull/41

llvm/utils/benchmark/src/benchmark_register.h
```diff
@@ -2,6 +2,7 @@
 #define BENCHMARK_REGISTER_H
 
 #include <vector>
+#include <limits>
```
bryanpkc, line 4: Nit: LLVM Coding Standard requires sorting the `#include` lines in alphabetical order. See the original patch b498303066. Consider cherry-picking the commit preserving original authorship.

michalpasztamobica: Good idea, I only realised that this has been fixed, when @pawosm-arm noticed that on `release_12x` the include is already there. I didn't expect that the commit would exactly match my fix :)

Should be good now, including the authorship preservation.

michalpasztamobica: Good idea, I only realised that this has been fixed, when @pawosm-arm noticed that on `release_12x` the include is already there. I didn't expect that the commit would exactly match my fix :)

Should be good now, including the authorship preservation.

bryanpkc (approved):
LGTM.

SouraVX (approved):

PR #45, opened 2021-05-21
[ClassicFlang][Driver] Correct the LLVM version passed by the Driver

SouraVX:

PR #47, opened 2021-05-23
[ClassicFlang][Driver] Correct the LLVM version passed by the Driver

SouraVX:
Essentially same as #45, just on release_12x branch.


michalpasztamobica (approved):
This looks good, especially if it works... ;-) I do have a question though - what if we update to 12.1? Should we then also bump this number here?

SouraVX:
> This looks good, especially if it works... ;-) I do have a question though - what if we update to 12.1? Should we then also bump this number here?

I think there aren't any major breaking changes across minor versions. We should be able to keep the same number.
However its a good question, I want to here @bryanpkc @kiranchandramohan thoughts on this.

kiranchandramohan (approved):
LGTM.

PR #48, opened 2021-06-04
Allow pre-compile jobs to be trigerred manually.

michalpasztamobica:
The `workflow_dispatch` must be visible in the `yml` script present in the repository's default branch, which currently is `release_100` but it will let us run the job on any other branch (`release_11x` or `release12x`).

### DIFF
--- a/.github/workflows/pre-compile_llvm.yml
+++ b/.github/workflows/pre-compile_llvm.yml
@@ -1,6 +1,7 @@
 name: Pre-compile llvm
 
 on:
+  workflow_dispatch:
   push:
     branches: [ release_100, release_11x ]
 


SouraVX (approved):
Thanks for the PR!
If I understood correctly i.e
From Slack:
> Once this is merged, if a repo admin/owner goes to Actions->pre-compile llvm, they will have a Dispatch button available and will need to manually run the job for release_11x and release_12x.

This will only be available to admins/owner ? I'm Okay with it :)
But any rationale behind it ?

pawosm-arm (approved):

michalpasztamobica:
@SouraVX , this is how Github Actions are configured by Github - there is no way we can influence it. I believe this is a security measure.

Personally, I think it is reasonable for repo owners to be able to somehow limit who and when is running the CI jobs. 
Obviously - opening PRs to a repo is available for anyone, but that's a tedious way of triggering CI ;) And you cannot rerun your own PR's jobs unless you push to the branch, which will automatically cancel the older run. I believe this is a protection from someone maliciously triggering builds to flood the servers with CI runs.
In our case this is particularly important as Arm has recently exposed their own ARM64 servers for PRs to use. If anyone was able to trigger, run or re-run jobs the servers could easily be taken out of the game effectively killing our CI. As long as it's just PR authors and repo owners - we are on the safe side :)

SouraVX:
@michalpasztamobica , thanks for detailed explanation :)

PR #58, opened 2021-09-09
Removing CI test-base for release_100 (release_100).

abrahamtovarmob:
This PR depends on https://github.com/flang-compiler/flang/pull/1133

### DIFF
--- a/.github/workflows/flang-tests.yml
+++ b/.github/workflows/flang-tests.yml
@@ -2,7 +2,7 @@ name: Build and test Flang
 
 on:
   pull_request:
-    branches: [ release_100, release_11x, release_12x ]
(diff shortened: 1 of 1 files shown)


bryanpkc:
I don't think this is necessary. Why don't we just leave it alone, in case we build the `release_100` branch for some reason in the future?

bryanpkc:
I think you should remove the trigger in `.github/workflows/flang-tests.yml` instead.

abrahamtovarmob:
Moved the change from the build to the test file

michalpasztamobica (approved):
This looks good to me :)

bryanpkc (approved):

PR #66, opened 2021-11-15
Pass target-attributes to Flang to embed in .ll files

RichBarton-Arm:
Have the clang driver pass target-attributes through to flang(2) so that they can be embedded in LLVM metadata.

The use case is LTO, where the target features need to be read in from the various inputs by the LTO plugin and merged. For flang, the non-LTO compilation works without this because the call to clang on the resulting .ll file from flang2 has the target-attributes passed to it by the flang driver. For the LTO case, these are need to already be present in the .ll file metadata so flang has to know about them.

I wanted to add a test case so I have include a commit to lit to enable a %flang lit substitution that lets us call it from lit in the clang/test repository. All tests here ought to be written as driver only tests (using `-###` or similar option) because flang1/flang2 are not necessarily available here (although they will be if they can be added to the test environment's $PATH). Added a simple test as proof of concept of this change in addition to the test for the target features.

See https://github.com/flang-compiler/flang/pull/1173 for the flang half.

### DIFF
--- a/clang/lib/Driver/ToolChains/Clang.cpp
+++ b/clang/lib/Driver/ToolChains/Clang.cpp
@@ -301,76 +301,11 @@ static void ParseMPreferVectorWidth(const Driver &D, const ArgList &Args,
   }
 }
 
-static void getWebAssemblyTargetFeatures(const ArgList &Args,
-                                         std::vector<StringRef> &Features) {
-  handleTargetFeaturesGroup(Args, Features, options::OPT_m_wasm_Features_Group);
-}
-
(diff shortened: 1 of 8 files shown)


RichBarton-Arm:
Added a second patch which cleans up the target feature string before sending to flang (an oversight in the original patch - my bad!)

RichBarton-Arm:
@bryanpkc  - have you had a chance to review this and compare with your own downstream implementation? Would be good to get this in upstream in a way that works for both of us. Same for the flang-side commit: https://github.com/flang-compiler/flang/pull/1173

pawosm-arm:
Covered by #84

PR #67, opened 2021-11-23
LLVM release_13x CI changes

abrahamtovarmob:
Changes to LLVM release_13x to be able to be build by github
Minor changes to code format to remove commented code and just disable it.

PR #68, opened 2021-12-06
Pass target-attributes to Flang to embed in .ll files / release_13x port

abrahamtovarmob:
Depends on https://github.com/flang-compiler/classic-flang-llvm-project/pull/67 to keep commit consistency

Adding same functionality to pass target-attributes to Flang to embed in .ll files as in https://github.com/flang-compiler/classic-flang-llvm-project/pull/66 but ported for release_13x


pawosm-arm:
This is also covered by PR #84

PR #69, opened 2021-12-08
Pass msve-vector-bits to Flang to embed in .ll files / release_13x

abrahamtovarmob:
Depends on https://github.com/flang-compiler/classic-flang-llvm-project/pull/67
Depends on https://github.com/flang-compiler/classic-flang-llvm-project/pull/68

Pass LLVM msve_vector_bits (simple string) to flang to embed in  generated .ll files. For normal compilation this won't make much  difference as the attributes are passed to clang after flang2 and can be applied there but this is crucial to enable LTO with flang  as clang will not apply the attributes when building the flang2 output. libLTO will need to read these out of the object files to apply them.


pawosm-arm:
Duplicated by PR #84

PR #70, opened 2021-12-10
LLVM release_13x CI changes

abrahamtovarmob:
Changes to LLVM release_13x to be able to be build by github
Minor fix to Docker recipe (fixing rpc error)


kiranchandramohan:
What is the difference compared to https://github.com/flang-compiler/classic-flang-llvm-project/pull/67?

abrahamtovarmob:
> What is the difference compared to #67?

Instead of using "apt" I'm using "apt-get".

kiranchandramohan:
Can you add that info to the commit message? Also this patch seems to include most of the previous commit as well. Not sure why it is not the incremental change of apt-get. Could you clarify?

abrahamtovarmob:
> 

As the fix was a line I didn't considered worthy enough to leave it in a standalone commit, so I squashed with his parent, for this reason git shows as I'm just forcing a change to the same commit. I can leave it as a standalone change so the fix can be clear. Please advice.

PR #72, opened 2021-12-10
Fix release_13x CI dockerfile

abrahamtovarmob:
Change to use "apt-get" instead, seems root has issues to understand "apt". Add -y to force to update without user intervention. I test it locally without issues.


kiranchandramohan (approved):
LGTM.

Merging since this is for fixing CI.

PR #73, opened 2021-12-10
Add platform independent build script for LLVM

tdusnoki:


scripts/build_llvm_project.py
```diff
@@ -0,0 +1,131 @@
+#!/usr/bin/env python
+
+import argparse
+import multiprocessing
```
kaadam: It seems to me you don't use multiprocessing lib, so I think you can remove it.

scripts/build_llvm_project.py
```diff
@@ -0,0 +1,131 @@
...
+                        help=f'specify build directory (default: {settings.LLVM_DIR}/%(default)s)')
+  buildopt.add_argument('--clean', action='store_true', default=False,
+                        help='clean build')
+  buildopt.add_argument('--build-type', metavar='TYPE', default='Release',
+                        help='set build type (default: %(default)s)')
+  buildopt.add_argument('--debug', dest='build_type', action='store_const', const='Debug', default=argparse.SUPPRESS,
+                        help='debug build (alias for --build-type %(const)s)')
+  buildopt.add_argument('--cmake-param', metavar='OPT', action='append', default=[],
+                        help='add custom argument to CMake')
+  buildopt.add_argument('--llvm-enable-projects', metavar='OPT', action='append', default=['clang'] if sys.platform == 'win32' else ['clang','openmp'],
```
kaadam: This line is a bit long, could you create a function to determine which project should be built based on the platform?

scripts/cmake/toolchain_windows_amd64.cmake
```diff
@@ -0,0 +1,2 @@
+set(CMAKE_C_COMPILER "d:/wrapper/install/bin/clang-cl.exe")
```
kaadam: Maybe you forgot to change this PATH.

scripts/build_llvm_project.py
```diff
@@ -0,0 +1,131 @@
...
+
+  return build_path
+
+def configure_llvm(arguments):
+  build_path = normalize_builddir(
+    settings.LLVM_DIR, arguments.builddir, arguments.clean)
+
+  build_options = generate_buildoptions(arguments)
+  install_root = Path(arguments.install_prefix)
+  additional_options = [
```
kaadam: I think these additional options could be moved into base_cmake_args.

scripts/build_llvm_project.py
```diff
@@ -0,0 +1,131 @@
...
+    f'-DLLVM_TARGETS_TO_BUILD={arguments.target}',
+    '-DLLVM_ENABLE_CLASSIC_FLANG=ON',
+    f'-DLLVM_ENABLE_PROJECTS={";".join(x for x in arguments.llvm_enable_projects)}'
+  ]
+  build_options.extend(additional_options)
+  cmake_cmd = ['cmake', '-B' + build_path, '-S' + os.path.join(settings.LLVM_DIR, 'llvm')]
```
kaadam: Just a nit: maybe LLVM_ROOT instead of LLVM_DIR? Or you could move this into setting and just define it there.
something like this: 
cmake_cmd = ['cmake', '-B', build_path, '-S', settings.LLVM_DIR]

bryanpkc:
@tdusnoki Could you rebase this PR?

bryanpkc:
@tdusnoki Did you intend to close this MR?

PR #75, opened 2021-12-13
[DebugInfo] Upgrade DISubragne::count to accept DIExpression also

alokkrsharma:
This PR is a backport from release_13x to release_11x (cherry-pick) of below.

```
commit 9fb0025f7084db14d16fe5cbe1d799f4ee0020af
Author: Alok Kumar Sharma <AlokKumar.Sharma@amd.com>
Date:   Thu Mar 25 16:34:57 2021 +0530
```
This should help better debug info, below are the details

This is needed for Fortran assumed shape arrays whose dimensions are
defined as,
  - 'count' is taken from array descriptor passed as parameter by
    caller, access from descriptor is defined by type DIExpression.
  - 'lowerBound' is defined by callee.
The current alternate way represents using upperBound in place of
count, where upperBound is calculated in callee in a temp variable
using lowerBound and count

Representation with count (DIExpression) is not only clearer as
compared to upperBound (DIVariable) but it has another advantage that
variable count is accessed by being parameter has better chance of
survival at higher optimization level than upperBound being local
variable.

Reviewed By: aprantl

Differential Revision: https://reviews.llvm.org/D99335

### DIFF
--- a/llvm/include/llvm/IR/DebugInfoMetadata.h
+++ b/llvm/include/llvm/IR/DebugInfoMetadata.h
@@ -336,10 +336,9 @@ public:
 
   Metadata *getRawStride() const { return getOperand(3).get(); }
 
-  typedef PointerUnion<ConstantInt*, DIVariable*> CountType;
   typedef PointerUnion<ConstantInt *, DIVariable *, DIExpression *> BoundType;
 
-  CountType getCount() const;
+  BoundType getCount() const;
 
   BoundType getLowerBound() const;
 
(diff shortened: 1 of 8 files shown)


bryanpkc (approved):
LGTM

bryanpkc:
Please rebase the branch and see if the macOS CI job passes.

alokkrsharma:
Hi @pawosm-arm , can you please approve this as well ?

pawosm-arm (approved):

PR #78, opened 2021-12-15
Introduction of gcc-11 to CI

pwisniewskimobica:
Switch to gcc11 and g++11

### DIFF
--- a/.github/workflows/pre-compile_llvm.yml
+++ b/.github/workflows/pre-compile_llvm.yml
@@ -3,7 +3,7 @@ name: Pre-compile llvm
 on:
   workflow_dispatch:
   push:
-    branches: [ release_100, release_11x ]
+    branches: [ release_100, release_11x, release_12x, release_13x]
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


pwisniewskimobica:
We want to introduce compilation with gcc-11 in Flang CI. 
This PR lets us build the gcc-11 artifacts for later use in Flang repository CI which is done in 
https://github.com/flang-compiler/flang/pull/1190

Kind request for your reviews, @bryanpkc , @shivaramaarao , @mleair , @kiranchandramohan
Thank you in advance!

.github/workflows/pre-compile_llvm.yml
```diff
@@ -35,18 +32,19 @@ jobs:
         run: |
           sudo add-apt-repository ppa:ubuntu-toolchain-r/ppa
```
bryanpkc, line 32: I think the logic would be easier to understand if you add the following here, and delete `env:` section above:
```
echo "cpp=g++" >> $GITHUB_ENV
```

kiranchandramohan:
I think release 10 is not used now. So you may close this.

.github/workflows/pre-compile_llvm.yml
pwisniewskimobica: fixed

kiranchandramohan:
Closing since we do not support release_10 now.

PR #80, opened 2021-12-15
Introduction of gcc-11 to CI 

pwisniewskimobica:
Switch to gcc11 and g++11

PR #81, opened 2021-12-15
Introduction of gcc-11 to CI

pwisniewskimobica:
Switch to gcc11 and g++11

### DIFF
--- a/.github/workflows/pre-compile_llvm.yml
+++ b/.github/workflows/pre-compile_llvm.yml
@@ -11,18 +11,13 @@ jobs:
     strategy:
       matrix:
         target: [X86]
-        cc: [clang]
-        cpp: [clang++]
+        cc: [clang, gcc]
         version: [10, 11]
-        include:
-          - target: X86
-            cc: gcc
-            cpp: g++
-            version: 10
             
     steps:
       - if: matrix.cc == 'clang'
         run: |
+          echo "cpp=clang++" >> $GITHUB_ENV
           sudo add-apt-repository ppa:ubuntu-toolchain-r/ppa
           sudo touch /etc/apt/sources.list.d/llvm.list
           echo 'deb http://apt.llvm.org/focal/ llvm-toolchain-focal-11 main' | sudo tee -a /etc/apt/sources.list.d/llvm.list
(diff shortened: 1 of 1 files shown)


pwisniewskimobica:
We want to introduce compilation with gcc-11 in Flang CI.
This PR lets us build the gcc-11 artifacts for later use in Flang repository CI which is done in
flang-compiler/flang#1190

Kind request for your reviews, @bryanpkc , @shivaramaarao , @mleair , @kiranchandramohan
Thank you in advance!

bryanpkc (approved):
LGTM

pawosm-arm:
What about the merge commit this PR contains? Are we happy with that @bryanpkc ?

bryanpkc (changes requested):
@pawosm-arm Thanks for catching the merge commit! I hadn't noticed it since the diffs looked identical to #80, and that one seemed fine from a commit perspective.

@pwisniewskimobica Could you please rebase this branch and remove the merge commit?

pwisniewskimobica:
@bryanpkc 
@pawosm-arm 
rebase applied
-->fixed

.github/workflows/pre-compile_llvm.yml
```diff
@@ -3,26 +3,23 @@ name: Pre-compile llvm
 on:
   workflow_dispatch:
   push:
-    branches: [ release_100, release_11x, release_12x, release_13x ]
+    branches: [ release_100, release_11x, release_12x, release_13x]
```
bryanpkc: Unnecessary diff. If you insist on removing the space, you should do the same after the left bracket, and also make sure that the formatting is the same in the other two PRs.

.github/workflows/pre-compile_llvm.yml
```diff
@@ -3,26 +3,23 @@ name: Pre-compile llvm
...
 
 jobs:
   build:
+    env:
+      cpp: "g++"
```
bryanpkc: Remove this and use `echo "cpp=g++" >> $GITHUB_ENV`in the appropriate steps below.

bryanpkc:
I would suggest rewording the commit to something like:
```
Use GCC 11 in CI builds
```

.github/workflows/pre-compile_llvm.yml
pwisniewskimobica: fixed

.github/workflows/pre-compile_llvm.yml
pwisniewskimobica: fixed

bryanpkc:
Could you remove the word "fix" in the commit message?

pwisniewskimobica:
@bryanpkc redundant word 'fix removed
-->fixed

pawosm-arm (approved):

PR #84, opened 2021-12-22
Pass msve-vector-bits to Flang to embed in .ll files / release_13x

pwisniewskimobica:
It is duplicate of https://github.com/flang-compiler/classic-flang-llvm-project/pull/69 but add more variants to test file.

Depends on #67
Depends on #68

Pass LLVM msve_vector_bits (simple string) to flang to embed in generated .ll files. For normal compilation this won't make much difference as the attributes are passed to clang after flang2 and can be applied there but this is crucial to enable LTO with flang as clang will not apply the attributes when building the flang2 output. libLTO will need to read these out of the object files to apply them.
For proper building classic llvm tests i had to add new option to build-llvm-project.sh to  decide which additional subprojects should be build. By default there are open mp and clang but for test purposes (this ticket) only clang is selected.

PR #86, opened 2022-01-24
Alternative implementation of pass msve-vector-bits to Flang to embed in .ll files / release_13x

pawosm-arm:
This is a new reimplementation of the PR #84

Corresponding change to the flang-compiler/flang project needs to be applied.

### DIFF
--- a/.github/workflows/clang-tests.yml
+++ b/.github/workflows/clang-tests.yml
@@ -18,9 +18,12 @@ jobs:
   build_clang:
     name: clang check-all
     runs-on: ${{ matrix.os }}
+    env:
+      install_prefix: /usr/local
     strategy:
       fail-fast: false
       matrix:
+        target: [X86]
         os:
           - ubuntu-latest
           - macOS-latest
@@ -39,8 +42,16 @@ jobs:
       uses: hendrikmuhs/ccache-action@v1
       with:
         key: ${{ matrix.os }}
-    - name: Test clang
-      uses: llvm/actions/build-test-llvm-project@main
-      with:
-        cmake_args: -G Ninja  -DLLVM_ENABLE_PROJECTS="clang" -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
-        build_target: check-clang
+    - name: Test clang macOS
+      if: ${{ matrix.os == 'macOS-latest'}}
+      run: |
+        nproc="sysctl -n hw.logicalcpu"
+        ./build-llvm-project.sh -t X86 -e "clang" -p /usr/local -s -n `$nproc` -i
+        cd build
+        make check-all
+    - name: Test clang ubuntu
+      if: ${{ matrix.os == 'ubuntu-latest'}}
+      run: |
+        ./build-llvm-project.sh -t X86 -p /usr/local -s -n $(nproc) -i
+        cd build
+        make check-all
--- a/clang/test/Driver/aarch64-implied-sve-features.c
+++ b/clang/test/Driver/aarch64-implied-sve-features.c
@@ -1,80 +1,80 @@
 // REQUIRES: aarch64-registered-target
 
-// RUN: %clang -target aarch64-linux-gnu -march=armv8-a+sve %s -### |& FileCheck %s --check-prefix=SVE-ONLY
+// RUN: %clang -target aarch64-linux-gnu -march=armv8-a+sve %s -### 2>&1 | FileCheck %s --check-prefix=SVE-ONLY
 // SVE-ONLY: "-target-feature" "+sve"
 
-// RUN: %clang -target aarch64-linux-gnu -march=armv8-a+nosve %s -### |& FileCheck %s --check-prefix=NOSVE
+// RUN: %clang -target aarch64-linux-gnu -march=armv8-a+nosve %s -### 2>&1 | FileCheck %s --check-prefix=NOSVE
 // NOSVE: "-target-feature" "-sve"
 
-// RUN: %clang -target aarch64-linux-gnu -march=armv8-a+sve+nosve %s -### |& FileCheck %s --check-prefix=SVE-REVERT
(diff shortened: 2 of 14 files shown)


.github/workflows/clang-tests.yml
```diff
@@ -39,8 +42,16 @@ jobs:
       uses: hendrikmuhs/ccache-action@v1
       with:
         key: ${{ matrix.os }}
-    - name: Test clang
-      uses: llvm/actions/build-test-llvm-project@main
-      with:
-        cmake_args: -G Ninja  -DLLVM_ENABLE_PROJECTS="clang" -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
-        build_target: check-clang
+    - name: Test clang macOs
```
bryanpkc: Why isn't the name of the operating system spelled the right way, i.e. `macOS`?

clang/test/Driver/flang/emit-flang-attrs.f90
```diff
@@ -0,0 +1,57 @@
+! REQUIRES: aarch64-registered-target
```
bryanpkc: Should this test be marked `REQUIRES: classic_flang`? Or will it work with LLVM Flang?

clang/test/Driver/flang/flang.f90
```diff
@@ -19,32 +19,32 @@
 
 ! Check that f90 files are not treated as "previously preprocessed"
 ! ... in --driver-mode=flang.
-! RUN: %clang --driver-mode=flang -### -E                  %s 2>&1 | FileCheck --check-prefixes=ALL,CHECK-E %s
+! RUN: %flang -### -E                  %s 2>&1 | FileCheck --check-prefixes=ALL,CHECK-E %s
```
bryanpkc: This test is marked `UNSUPPORTED: classic_flang`. Did you verify that this will work correctly with LLVM Flang?

bryanpkc:
Can the commits 222b18ff15c1445ef07cc79551e90d83d1225933, 84262af9e398518d0792c9e213ffe5295c762286, 5a9a2229f8274dcc780ecf73a1be0af3ac63a424 be upstreamed in a form that works with both Classic Flang and LLVM Flang?

clang/test/Driver/flang/emit-flang-attrs.f90
pawosm-arm: > Can the commits [222b18f](https://github.com/flang-compiler/classic-flang-llvm-project/commit/222b18ff15c1445ef07cc79551e90d83d1225933), [84262af](https://github.com/flang-compiler/classic-flang-llvm-project/commit/84262af9e398518d0792c9e213ffe5295c762286), [5a9a222](https://github.com/flang-compiler/classic-flang-llvm-project/commit/5a9a2229f8274dcc780ecf73a1be0af3ac63a424) be upstreamed in a form that works with both Classic Flang and LLVM Flang?

No, at least, not by us.

pawosm-arm:
@bryanpkc those tests (with `REQUIRES: aarch64-registered-target`) specifically require AArch64 targets as they refer to AArch64-specific features. We have no plans to push new-flang related things to classic-flang-llvm-project similarly as we we have no plans to publish changes that would ensure such compatibility to already submitted test cases.

.github/workflows/clang-tests.yml
pawosm-arm: a typo. fixed.

clang/test/Driver/flang/flang.f90
pawosm-arm: No, I don't have any workable LLVM Flang build at my disposal.

bryanpkc:
@pawosm-arm I understand your position. My question is, why is this PR modifying tests that are only exercised for the new Flang? They will not run in a Classic Flang build.

pawosm-arm:
> @pawosm-arm I understand your position. My question is, why is this PR modifying tests that are only exercised for the new Flang? They will not run in a Classic Flang build.

I do understand your point now. Thing is, my part of this PR is just the core logic, the rest I have inherited from other PR's that my code required to be merged to make it fully working. I've reached the author and we agreed that the easiest is to just remove the parts that are not necessary for my code to work. Also, the test that this commit extends should not be moved to the `flang` directory as it is not being executed for the classic flang when it is there.

pawosm-arm:
Hello all,
what is the current reviewer's standpoint on this one?

RichBarton-Arm:
If I am interpreting Bryan's request correctly, I think we need to add `REQUIRES: classic_flang` on all the new testcases added so that they only run when the codebase is built with `ENABLE_CLASSIC_FLANG` is set. They perhaps should also live in the `flang` subdirectory of `clang/test/Driver`. This is a mistake in my original patch that you have inherited here.

Thinking aloud. Perhaps it would be less confusing to put all the classic flang tests into a new subdirectory called `classic_flang`. We could then apply the `REQUIRES`/`UNSUPPORTED` at a top level on each directory rather than in each test individually? One for a future patch though.

clang/test/Driver/aarch64-implied-sve-features.c
```diff
@@ -0,0 +1,78 @@
+// RUN: %clang -target aarch64-linux-gnu -march=armv8-a+sve %s -### 2>&1 | FileCheck %s --check-prefix=SVE-ONLY
```
bryanpkc, line 3: Does this test need `REQUIRES: aarch64-registered-target`?

bryanpkc:
I like @RichBarton-Arm's suggestion.

clang/test/Driver/aarch64-implied-sve-features.c
pawosm-arm: Proposed change makes sense, and I'll add it to this PR, but bear in mind, I'm pulling in here an mainline llvm-project commit.

pawosm-arm:
> I like @RichBarton-Arm's suggestion.

It is a good suggestion, but I think it goes way beyond the scope of this PR. Meanwhile, I've added missing `REQUIRES: classic_flang` directives.

bryanpkc (approved):
LGTM.

PR #90, opened 2022-02-15
Introduce handling of the -dM option into flang

pwisniewskimobica:
It was decided to add clang -dM option to flang.
It prints out all the macros defined - similarly as clang or gfortran does it.
It works only with flang PR for this: https://github.com/flang-compiler/flang/pull/1223

### DIFF
--- a/clang/lib/Driver/ToolChains/ClassicFlang.cpp
+++ b/clang/lib/Driver/ToolChains/ClassicFlang.cpp
@@ -132,6 +132,11 @@ void ClassicFlang::ConstructJob(Compilation &C, const JobAction &JA,
     CommonCmdArgs.push_back("-nosave");
   }
 
+  for (auto Arg : Args.filtered(options::OPT_dM)) {
+    Arg->claim();
+    UpperCmdArgs.push_back("-list-macros");
+  }
+
   // Treat denormalized numbers as zero: On
   for (auto Arg : Args.filtered(options::OPT_Mdaz_on)) {
     Arg->claim();
--- a/clang/test/Driver/flang/list-macros.f95
+++ b/clang/test/Driver/flang/list-macros.f95
@@ -1,0 +1,8 @@
+! REQUIRES: classic_flang
+
+! Check that the driver invokes flang1 correctly when flang invokes with -dM option
+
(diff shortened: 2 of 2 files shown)


pawosm-arm:
Please use more words in the commit message. They can really be multi-word multi-line bits of a text. Would be great if the commit message could mention that it requires corresponding commit to be applied on the classic flang repository.
Also please avoid using underscores in the commit messages. UNIX shell allows you to pass multi-word parameters to the commands, e.g.:
```
$ git commit file.c -s -m "This commit refactors file.c"
```

pawosm-arm (changes requested):
Use more words in the commit message.

clang/lib/Driver/ToolChains/ClassicFlang.cpp
```diff
@@ -132,6 +132,11 @@ void ClassicFlang::ConstructJob(Compilation &C, const JobAction &JA,
     CommonCmdArgs.push_back("-nosave");
   }
 
+  for (auto Arg : Args.filtered(options::OPT_dM)) {
+    Arg->claim();
+    CommonCmdArgs.push_back("-list-macros");
```
bryanpkc: Do we need to pass `-list-macros` to `flang2`? Or just `flang1`?

pwisniewskimobica: @bryanpkc 
-list-macros is passed just to flang1 and this is enough

bryanpkc: If `-list-macros` is only needed for `flang1`, then you should push it to `UpperCmdArgs` instead of `CommonCmdArgs`.

pwisniewskimobica: @bryanpkc changed to `UpperCmdArgs`

bryanpkc (approved):
LGTM.

bryanpkc:
Actually it would be a good idea to add a simple test for this option under `clang/test/Driver/flang/`.

pwisniewskimobica:
@bryanpkc test added (list-macros.f95)

pwisniewskimobica:
@pawosm-arm All remarks applied.Please look again.Thank you

pawosm-arm (approved):

shivaramaarao (approved):

PR #94, opened 2022-03-07
Allow /dev/null to be an input file.

pwisniewskimobica:
This is patch for following problem:
`Problem is in flang that when -</dev/null is putted as an input file, then the clang is invoked instead of flang1.
We can see this via invoking example command:
flang -### -E -</dev/null
Then the clang is raised in first order.
We would like to to behave similar as we would invoke:
flang -### -E <input_file.f90>
In this scenario we can see that flang1 is raised in first order`

Make fortran mode when an input is '-' character indicating stdin.
Make temporary file and make /dev/stdin an input to flang1.
Test file for this change also is added

@bryanpkc @kiranchandramohan @michalpasztamobica @pawosm-arm @RichBarton-Arm @shivaramaarao Please take a look on the patch.Thanks

### DIFF
--- a/clang/lib/Driver/Driver.cpp
+++ b/clang/lib/Driver/Driver.cpp
@@ -2187,7 +2187,14 @@ void Driver::BuildInputs(const ToolChain &TC, DerivedArgList &Args,
           if (!Args.hasArgNoClaim(options::OPT_E) && !CCCIsCPP())
             Diag(IsCLMode() ? clang::diag::err_drv_unknown_stdin_type_clang_cl
                             : clang::diag::err_drv_unknown_stdin_type);
-          Ty = types::TY_C;
+          if (IsFlangMode()) { 
+#ifdef ENABLE_CLASSIC_FLANG
+            Ty = types::TY_PP_F_FreeForm; 
+#else
+            Ty = types::TY_PP_Fortran;
+#endif
+          } else
+            Ty = types::TY_C;
         } else {
           // Otherwise lookup by extension.
           // Fallback is C if invoked as C preprocessor, C++ if invoked with
--- a/clang/lib/Driver/ToolChains/ClassicFlang.cpp
+++ b/clang/lib/Driver/ToolChains/ClassicFlang.cpp
@@ -82,12 +82,23 @@ void ClassicFlang::ConstructJob(Compilation &C, const JobAction &JA,
     Stem = C.getDriver().GetTemporaryPath("", "");
   } else {
     OutFile = Output.getFilename();
+    if (OutFile == "-.ll")
+      OutFile = "tmp";
     Stem = llvm::sys::path::filename(OutFile);
     llvm::sys::path::replace_extension(Stem, "");
   }
 
   // Add input file name to the compilation line
-  UpperCmdArgs.push_back(Input.getBaseInput());
+  const char *InputFile = Input.getBaseInput();
+  if (strcmp(InputFile, "-") == 0)
+  // FIXME: done specially for handling flang -E -dM - < /dev/null
+#if defined(_WIN32) || defined(_WIN64)
+    UpperCmdArgs.push_back("nul");
+#else
+    UpperCmdArgs.push_back("/dev/stdin");
+#endif
+  else
+    UpperCmdArgs.push_back(Input.getBaseInput());
 
   // Add temporary output for ILM
   const char * ILMFile = Args.MakeArgString(Stem + ".ilm");
--- a/clang/test/Driver/flang/dev_null_as_input_windows.f95
+++ b/clang/test/Driver/flang/dev_null_as_input_windows.f95
@@ -1,0 +1,11 @@
+! REQUIRES: classic_flang
+! REQUIRES: system-windows
+
+! Check that the driver invokes flang1 correctly for preprocessed fixed-form
+! Fortran code.
+
+! RUN: %clang --driver-mode=flang -E -target x86_64-unknown-windows-gnu -c - < NUL -### 2>&1 \
+! RUN:   | FileCheck %s
+! CHECK: "{{.*}}flang1"
+! CHECK: "nul"
+! CHECK: "tmp.ilm"
--- a/clang/test/Driver/flang/dev_null_as_input_linux.f95
+++ b/clang/test/Driver/flang/dev_null_as_input_linux.f95
@@ -1,0 +1,11 @@
+! REQUIRES: classic_flang
+! REQUIRES: system-linux
+
+! Check that the driver invokes flang1 correctly for preprocessed fixed-form
+! Fortran code.
+
+! RUN: %clang --driver-mode=flang -E -target x86_64-unknown-linux-gnu -c - </dev/null -### 2>&1 \
+! RUN:   | FileCheck %s
+! CHECK: "{{.*}}flang1"
+! CHECK: "/dev/stdin"
+! CHECK: "tmp.ilm"


clang/lib/Driver/Driver.cpp
```diff
@@ -2187,7 +2187,17 @@ void Driver::BuildInputs(const ToolChain &TC, DerivedArgList &Args,
           if (!Args.hasArgNoClaim(options::OPT_E) && !CCCIsCPP())
             Diag(IsCLMode() ? clang::diag::err_drv_unknown_stdin_type_clang_cl
                             : clang::diag::err_drv_unknown_stdin_type);
-          Ty = types::TY_C;
+          if (memcmp(Value, "-", 2) == 0) { 
+            if (IsFlangMode()) { 
+#ifdef ENABLE_CLASSIC_FLANG
+              Ty = types::TY_PP_F_FreeForm; 
+#else
+              Ty = types:TY_PP_Fortran;
```
pawosm-arm: one `:` missing here... `types::TY_` not `types:TY_`

clang/lib/Driver/Driver.cpp
```diff
@@ -2187,7 +2187,17 @@ void Driver::BuildInputs(const ToolChain &TC, DerivedArgList &Args,
...
+#endif
+            }
+            else
```
pawosm-arm: move `else` to the line above, namely, `} else`

clang/lib/Driver/ToolChains/ClassicFlang.cpp
```diff
@@ -82,12 +82,18 @@ void ClassicFlang::ConstructJob(Compilation &C, const JobAction &JA,
...
     llvm::sys::path::replace_extension(Stem, "");
   }
 
   // Add input file name to the compilation line
-  UpperCmdArgs.push_back(Input.getBaseInput());
+  const char *InputFile = Input.getBaseInput();
+  if (strcmp(InputFile, "-") == 0)
+    UpperCmdArgs.push_back("/dev/stdin");
+  else
+   UpperCmdArgs.push_back(Input.getBaseInput());
```
pawosm-arm: wrong indentation. add one more space.

clang/test/Driver/flang/dev_null_as_input.f95
```diff
@@ -0,0 +1,10 @@
+! REQUIRES: classic_flang
+
+! Check that the driver invokes flang1 correctly for preprocessed fixed-form
+! Fortran code.
+
+! RUN: %clang --driver-mode=flang -E -target x86_64-unknown-linux-gnu -c -</dev/null -### 2>&1 \
```
pawosm-arm: this is OK, although it is hard to read. Could you put one space between `-` and `<`? Namely `-c - </dev/null`

pawosm-arm (changes requested):

pwisniewskimobica:
@pawosm-arm Thank You for remarks. All of them applied.

clang/lib/Driver/Driver.cpp
```diff
@@ -2187,7 +2187,16 @@ void Driver::BuildInputs(const ToolChain &TC, DerivedArgList &Args,
           if (!Args.hasArgNoClaim(options::OPT_E) && !CCCIsCPP())
             Diag(IsCLMode() ? clang::diag::err_drv_unknown_stdin_type_clang_cl
                             : clang::diag::err_drv_unknown_stdin_type);
-          Ty = types::TY_C;
+          if (memcmp(Value, "-", 2) == 0) {
```
pawosm-arm: I don't get it. Considering the `if` condition above (line 2181), is this condition ever false?

pwisniewskimobica: @pawosm-arm This condition is false when a real file is given to the input (not stdin).
Btw- this condition was done before- i did not make it.

pwisniewskimobica: @pawosm-arm now i got it. sorry. You are right.

pwisniewskimobica: the code is fixed. The `if (memcmp(Value, "-", 2) == 0)` condition was doubled.
Now one of them is removed.Please take a look.Thanks in advance

pawosm-arm (approved):

clang/lib/Driver/ToolChains/ClassicFlang.cpp
```diff
@@ -82,12 +82,18 @@ void ClassicFlang::ConstructJob(Compilation &C, const JobAction &JA,
...
+      OutFile = "tmp";
     Stem = llvm::sys::path::filename(OutFile);
     llvm::sys::path::replace_extension(Stem, "");
   }
 
   // Add input file name to the compilation line
-  UpperCmdArgs.push_back(Input.getBaseInput());
+  const char *InputFile = Input.getBaseInput();
+  if (strcmp(InputFile, "-") == 0)
+    UpperCmdArgs.push_back("/dev/stdin");
```
bryanpkc, line 98: I guess this wouldn't work on Windows? @kaadam @xoviat @isuruf Any thought?

clang/test/Driver/flang/dev_null_as_input.f95
```diff
@@ -0,0 +1,10 @@
+! REQUIRES: classic_flang
+
+! Check that the driver invokes flang1 correctly for preprocessed fixed-form
+! Fortran code.
+
+! RUN: %clang --driver-mode=flang -E -target x86_64-unknown-linux-gnu -c - </dev/null -### 2>&1 \
```
bryanpkc: This test may need `REQUIRES: linux` for the use of `/dev/null` in the `RUN` command, in which case we probably should also have a Windows version of the test.

pwisniewskimobica: @bryanpkc What do You mean? Could you give an example?

clang/lib/Driver/ToolChains/ClassicFlang.cpp
pwisniewskimobica: @bryanpkc Since i cannot test it on windows have You got any idea how to do it there?

pawosm-arm: You can't do it on Windows. At least not with the 'path to a device file' approach, since Windows does not share the UNIX-specific 'everything is a file' philosophy. I guess the easiest way is to add the following preprocessor check:
```
#if !defined(_WIN32) && !defined(_WIN64)
...
#endif
```

bryanpkc: According to [StackOverflow](https://stackoverflow.com/questions/313111/is-there-a-dev-null-on-windows), on Windows, Command Prompt recognizes a pseudo-file named `NUL`, and PowerShell recognizes `$null`. While they work as the output of a pipe, I don't know whether they can be used as an input to a pipe.

clang/test/Driver/flang/dev_null_as_input.f95
bryanpkc: I meant that you should probably tell `llvm-lit` not to run this test if the current operating system is not Linux. But I think the correct feature tag to use is `system-linux`, not `linux`. Some examples are:

- https://github.com/llvm/llvm-project/blob/main/clang/test/Analysis/ctu-on-demand-parsing.c#L22
- https://github.com/llvm/llvm-project/blob/main/clang/test/Driver/amdgpu-openmp-system-arch.c#L1

pwisniewskimobica:
@pawosm-arm @bryanpkc 
I just made patch commit in this PR and in it I changed dev/stdin for simply creating and passing temp.f90 file.
I just checked it on Linux and its working. Unfortunately i have not got windows to test it, but i think it is more generous solution than creating separate version for windows and linux.
What You think?

bryanpkc (changes requested):
The latest change (simply calling `fopen` with a relative path name) risks overwriting a user's file, causing data loss. I also believe that LLVM already provides standard utility functions for creating temporary files.

However, instead of creating a temporary file, I think the correct way to do this is to pass the special filename `NUL` to `flang1`, but we'll have to verify whether that gives the proper behaviour on Windows.

Taking a step back, I think it would be a good idea to update the driver and `flang1` to allow them to work correctly (with or without `-E` and `-dM`) with Fortran code being streamed in via standard input, just like regular Clang.

pawosm-arm:
> The latest change (simply calling `fopen` with a relative path name) risks overwriting a user's file, causing data loss. I also believe that LLVM already provides standard utility functions for creating temporary files.
> 
> However, instead of creating a temporary file, I think the correct way to do this is to pass the special filename `NUL` to `flang1`, but we'll have to verify whether that gives the proper behaviour on Windows.
> 
> Taking a step back, I think it would be a good idea to update the driver and `flang1` to allow them to work correctly (with or without `-E` and `-dM`) with Fortran code being streamed in via standard input, just like regular Clang.

Curiously, I've found that the issue must have been faced in clang before by the others. See the `HIP.cpp` file in clang, it contains this:
```
#if defined(_WIN32) || defined(_WIN64)
#define NULL_FILE "nul"
#else
#define NULL_FILE "/dev/null"
#endif
```

pwisniewskimobica:
@bryanpkc @pawosm-arm Just made the same as it is in HIP.cpp.
Please take a look-
Thanks

clang/lib/Driver/ToolChains/ClassicFlang.cpp
```diff
@@ -91,7 +97,7 @@ void ClassicFlang::ConstructJob(Compilation &C, const JobAction &JA,
   // Add input file name to the compilation line
   const char *InputFile = Input.getBaseInput();
   if (strcmp(InputFile, "-") == 0)
-    UpperCmdArgs.push_back("/dev/stdin");
+    UpperCmdArgs.push_back(NULL_FILE);
```
pawosm-arm: You've just changed `/dev/stdin` to `/dev/null`....
The `HIP.cpp` I have mentioned in one of my commends was just an example how problems with Windows are handled in other parts of Clang, I did not mention that this is the final solution.

For me, this code should look like (assuming NULL_FILE definition you've added above is kept):
```
#if defined(_WIN32) || defined(_WIN64)
  UpperCmdArgs.push_back(NULL_FILE);
#else
  UpperCmdArgs.push_back("/dev/stdin");
#endif
```

pawosm-arm (changes requested):

clang/lib/Driver/ToolChains/ClassicFlang.cpp
pwisniewskimobica: @pawosm-arm having NULL_FILE definition and the piece of code You just gave have no much sense. Why doubling `#if defined(_WIN32) || defined(_WIN64)`?

pawosm-arm: it makes more sense than replacing stdin with null (for non-Windows case).

If you want to avoid doubling, try this instead:
```
#if defined(_WIN32) || defined(_WIN64)
  UpperCmdArgs.push_back("nul");
#else
  UpperCmdArgs.push_back("/dev/stdin");
#endif
```
...then you can remove previous NULL_FILE definition, won't be needed anymore.

pwisniewskimobica: @pawosm-arm done

clang/test/Driver/flang/dev_null_as_input.f95
pawosm-arm: @pwisniewskimobica this still needs to be changed.

pwisniewskimobica: @pawosm-arm 
Just fixed. Thanks

pawosm-arm (approved):

clang/test/Driver/flang/dev_null_as_input.f95
bryanpkc: Since there is now Windows support for the same feature, please provide a Windows version of this test.

pwisniewskimobica: @bryanpkc Instead of one test file I just created separate for windows and linux.
Unfortunately i dont have windows to test it so I am predicting how test should work. 
 Please take a look. Thanks

clang/test/Driver/flang/dev_null_as_input_windows.f95
```diff
@@ -0,0 +1,11 @@
+! REQUIRES: classic_flang
+! REQUIRES: system-windows
+
+! Check that the driver invokes flang1 correctly for preprocessed fixed-form
+! Fortran code.
+
+! RUN: %clang --driver-mode=flang -E -target x86_64-unknown-windows-gnu -c - </dev/null -### 2>&1 \
```
pawosm-arm: you can't use /dev/null for windows test...

pwisniewskimobica: @pawosm-arm I just changed it from `-</dev/null` to `-<%s`
Since I am working on linux i cannot double check its working but in my opinion it should because it should have the same results.
Please take a look. Thanks

pawosm-arm: yeah, it's probably OK.
@bryanpkc what do you think about that?

bryanpkc: You should be able to write `- < NUL` on Windows.

pwisniewskimobica: @bryanpkc Just changed it to Your recommendation. Thanks

bryanpkc:
Why is CI not being run on this PR?

*Update: It seems a PR in "Changes Requested" state wouldn't trigger re-runs on updates.*

clang/lib/Driver/ToolChains/ClassicFlang.cpp
```diff
@@ -82,12 +82,22 @@ void ClassicFlang::ConstructJob(Compilation &C, const JobAction &JA,
...
+      OutFile = "tmp";
     Stem = llvm::sys::path::filename(OutFile);
     llvm::sys::path::replace_extension(Stem, "");
   }
 
   // Add input file name to the compilation line
-  UpperCmdArgs.push_back(Input.getBaseInput());
+  const char *InputFile = Input.getBaseInput();
+  if (strcmp(InputFile, "-") == 0)
+#if defined(_WIN32) || defined(_WIN64)
```
bryanpkc, line 95: I think there should be a `FIXME` comment here to explain that treating the standard input as the null input is not correct, but we need this stop-gap solution to accommodate the common idiom of `flang -E -dM - < /dev/null`.

bryanpkc (approved):

clang/lib/Driver/ToolChains/ClassicFlang.cpp
pwisniewskimobica: @bryanpkc Just added comment
`//FIXME:done specially for handling flang -E -dM - < /dev/null`

clang/lib/Driver/ToolChains/ClassicFlang.cpp
```diff
@@ -82,12 +82,23 @@ void ClassicFlang::ConstructJob(Compilation &C, const JobAction &JA,
...
+      OutFile = "tmp";
     Stem = llvm::sys::path::filename(OutFile);
     llvm::sys::path::replace_extension(Stem, "");
   }
 
   // Add input file name to the compilation line
-  UpperCmdArgs.push_back(Input.getBaseInput());
+  const char *InputFile = Input.getBaseInput();
+  if (strcmp(InputFile, "-") == 0)
+  //FIXME:done specially for handling flang -E -dM - < /dev/null
```
pawosm-arm: a small nit: all comments in this file have one space after `//`, e.g. `// FIXME:...`

pwisniewskimobica: @pawosm-arm fixed. Thanks

bryanpkc:
@kaadam @tdusnoki @isuruf @xoviat Can one of you confirm whether the Windows test case in this PR works as intended?

kaadam:
> @kaadam @tdusnoki @isuruf @xoviat Can one of you confirm whether the Windows test case in this PR works as intended?

Sorry for the late response. I look at this later afternoon.

shivaramaarao (approved):

kiranchandramohan:
@kaadam Can you have a look soon? We would like to merge this tomorrow.

kaadam (approved):
Apologies for the delay in responding.

LGTM, the test invokes flang1, 'nul' and 'tmp.ilm' are in the input of flang1 as expected. 
Tested on X64 Windows machine.

PR #101, opened 2022-03-22
Canonicalize ENABLE_CLASSIC_FLANG for lit

RichBarton-Arm:
These CMake variables are used to set python variables in the lit
config. As such, they need to be set with a valid CMake boolean value
that is also a valid python boolean value (1, True, etc.) and not one
that is not (ON, NO, etc.) to work. This is fragile.

Call the LLVM cmake helper function on these two downstream CMake


RichBarton-Arm:
Adding reviewers.

pawosm-arm (approved):

bryanpkc (approved):

PR #103, opened 2022-03-22
Fix classic flang version screen

RichBarton-Arm:
Version screen has been erroneously reporting as flang-new since LLVM 12


bryanpkc (approved):
Thanks Rich. LGTM.

RichBarton-Arm:
Adding some more reviewers to get the second approval as per process.

pawosm-arm (approved):

PR #106, opened 2022-04-24
Port Classic Flang to LLVM 14

bryanpkc:
This PR ports Classic Flang-related changes from [flang-13.0.0-20220325](https://github.com/flang-compiler/classic-flang-llvm-project/commit/0254365051067b50372914150a004048c5cbce54) to [llvmorg-14.0.1](https://github.com/flang-compiler/classic-flang-llvm-project/commit/c62053979489ccb002efe411c3af059addcb5d7d), on which our new `release_14x` branch will be based.

I have squashed some commits together to resolve merge conflicts and make them buildable; in such cases, the original commits are cited. In some cases, I have undone the squashes performed by @abrahamtovarmob and separated commits that have clearly different purposes. Some upstream commits which were backported to `release_13x` did not need to be ported to this branch (89065d4abfe506af1da432ddf8651f1893a4a4fd, b4723b90ebe016fd0973c4afeffd97f0d7c3b66a, 0254365051067b50372914150a004048c5cbce54).

A commit was added on top of the branch to enable GitHub builds of the `release_14x` branch.


clang/include/clang/Driver/Options.td
```diff
@@ -4563,6 +4565,12 @@ defm devirtualize : BooleanFFlag<"devirtualize">, Group<clang_ignored_gcc_optimi
 defm devirtualize_speculatively : BooleanFFlag<"devirtualize-speculatively">,
     Group<clang_ignored_gcc_optimization_f_Group>;
 
+// gfortran options that we recognize in the driver and pass along when
```
RichBarton-Arm: I think this comment is supposed to live with the declaration of gfortran_Group? Doesn't seem to match the intention of these classic flang option groups?

clang/include/clang/Driver/Options.td
```diff
@@ -4595,13 +4605,19 @@ defm check_array_temporaries : BooleanFFlag<"check-array-temporaries">, Group<gf
...
 defm dump_fortran_original : BooleanFFlag<"dump-fortran-original">, Group<gfortran_Group>;
 defm dump_parse_tree : BooleanFFlag<"dump-parse-tree">, Group<gfortran_Group>;
 defm external_blas : BooleanFFlag<"external-blas">, Group<gfortran_Group>;
 defm f2c : BooleanFFlag<"f2c">, Group<gfortran_Group>;
-defm frontend_optimize : BooleanFFlag<"frontend-optimize">, Group<gfortran_Group>;
+def fixed_form_off : Flag<["-"], "fno-fixed-form">, Group<fortran_format_Group>,
+  HelpText<"Disable fixed-form format for Fortran">;
+def free_form_off : Flag<["-"], "fno-free-form">, Group<fortran_format_Group>,
+  HelpText<"Disable free-form format for Fortran">;
+defm frontend_optimize : BooleanFFlag<"frontend-optimize">, Group<fortran_format_Group>;
```
RichBarton-Arm: frontend_optimize is in the wrong group I think. Can stay in the gfortran_Group

clang/include/clang/Driver/Options.td
```diff
@@ -4644,56 +4660,77 @@ def Xflang : Separate<["-"], "Xflang">,
...
   HelpText<"Put MODULE files in <dir>">,
   DocBrief<[{This option specifies where to put .mod files for compiled modules.
 It is also added to the list of directories to be searched by an USE statement.
 The default is the current directory.}]>;
 
-def ffixed_form : Flag<["-"], "ffixed-form">, Group<f_Group>,
+def ffixed_form : Flag<["-"], "ffixed-form">, Group<fortran_format_Group>,
   HelpText<"Process source files in fixed form">;
-def ffree_form : Flag<["-"], "ffree-form">, Group<f_Group>,
+#ifdef ENABLE_CLASSIC_FLANG
```
RichBarton-Arm: I'm concerned about using ENABLE_CLASSIC_FLANG to guard additions to Options.td when it does not also guard alterations to Options.td. That seems confusing to me. I don't think there is any chance of an unintended downside to having classic flang options defined in the compiled opt table when classic flang is not included? They simply won't be used by the rest of the code.

clang/include/clang/Driver/Options.td
```diff
@@ -4644,56 +4660,77 @@ def Xflang : Separate<["-"], "Xflang">,
...
 defm xor_operator : OptInFC1FFlag<"xor-operator", "Enable .XOR. as a synonym of .NEQV.">;
 defm logical_abbreviations : OptInFC1FFlag<"logical-abbreviations", "Enable logical abbreviations">;
 defm implicit_none : OptInFC1FFlag<"implicit-none", "No implicit typing allowed unless overridden by IMPLICIT statements">;
 
 def fno_automatic : Flag<["-"], "fno-automatic">, Group<f_Group>,
   HelpText<"Implies the SAVE attribute for non-automatic local objects in subprograms unless RECURSIVE">;
+#endif
 }
 
+#ifdef ENABLE_CLASSIC_FLANG
```
RichBarton-Arm, line 4726: Although this is quite nice and clear, despite my comment above about not liking ENABLE_CLASSIC_FLANG use like this....

Perhaps I could be persuaded, but IMO we should just modify this file for classic flang without using ENABLE_CLASSIC_FLANG. What do others think?

clang/lib/Driver/Driver.cpp
```diff
@@ -4621,6 +4647,10 @@ class ToolSelector final {
     if (!T->hasIntegratedBackend() && !(OutputIsLLVM && T->canEmitIR()))
       return nullptr;
 
+    // Classic Flang is not integrated with the backend.
```
RichBarton-Arm, line 4655: Need to pull in https://github.com/flang-compiler/classic-flang-llvm-project/pull/100/files

RichBarton-Arm (commented):
I have recently done this exercise downstream so I took a look. I left a few comments in passing for minor things.

It looks like you need to cherry pick across my three recent PR #100 , #102  and #104 .

When doing this downstream, I found a regression in the way classic flang handles preprocessing with `-save-temps` (and some other not important options). `flang -E -save-temps file.F95` causes the preprocessor to be run twice. I have a fix so perhaps it is easiest if I make a separate PR and we can commit it over the top of this one?

bryanpkc:
I added a patch to revert an ancient change in LoopVectorize.cpp, which was causing an AArch64 test to fail. This probably needs a review by @vjayathirtha-nv or someone more familiar with the motivation of the original patch.

bryanpkc:
@RichBarton-Arm  Thanks for the review; I'll address your concerns soon.

bryanpkc:
> When doing this downstream, I found a regression in the way classic flang handles preprocessing with `-save-temps` (and some other not important options). `flang -E -save-temps file.F95` causes the preprocessor to be run twice. I have a fix so perhaps it is easiest if I make a separate PR and we can commit it over the top of this one?

That sounds like a good idea. In this PR I was only porting the basic patches from `flang-13.0.0-20220325` so there would have been a follow-up PR to cherry-pick other more recent patches. I also didn't notice the double preprocessing problem in our downstream compiler previously.

clang/include/clang/Driver/Options.td
bryanpkc: I agree with you. Simply porting the old inconsistent guards to LLVM 14 is easy, but we really ought to make these things consistent. I will take a deeper look at the uses of `ENABLE_CLASSIC_FLANG` in this file. This is a good opportunity to make the file compatible with both LLVM Flang and Classic Flang (we are playing with both downstream).

vjayathirtha-nv:
> I added a patch to revert an ancient change in LoopVectorize.cpp, which was causing an AArch64 test to fail. This probably needs a review by @vjayathirtha-nv or someone more familiar with the motivation of the original patch.

Hi @bryanpkc, [this comment](https://github.com/flang-compiler/llvm/commit/68669092a5ad1fd29f62346dc30973f86cfacbfe#commitcomment-32754393) describes the context for this change. However I'm not in charge of maintaining classic flang anymore, maybe @sscalpone or @schweitzpgi can help you further.

clang/lib/Driver/ToolChains/ClassicFlang.cpp
```diff
@@ -327,28 +327,34 @@ void ClassicFlang::ConstructJob(Compilation &C, const JobAction &JA,
...
-  }
 
-  // -gdwarf-3
-  for (auto Arg : Args.filtered(options::OPT_gdwarf_3)) {
-    Arg->claim();
-    CommonCmdArgs.push_back("-x");
-    CommonCmdArgs.push_back("120");
-    CommonCmdArgs.push_back("0x4000");
+    if (!GDwarfArg) // -g without -gdwarf-X produces default (DWARFv4)
+      CommonCmdArgs.push_back("0x1000000");
```
guy-david, line 358: Clang has moved to DWARF 5 by default, should this be updated as well?

bryanpkc: The option `-Mx,120,0x2000000` should enable DWARF 5, but I am not sure how well it is supported; more thorough testing and perhaps some additional fixes might be needed. I will create an issue to track this work.

clang/lib/Driver/Driver.cpp
bryanpkc: Done.

clang/include/clang/Driver/Options.td
bryanpkc: I have updated the uses of the macro in the file so that the conflicting Fortran options are better guarded.

clang/include/clang/Driver/Options.td
bryanpkc: Fixed.

clang/include/clang/Driver/Options.td
bryanpkc: The comment has been rewritten, and the option groups have been moved to be closer to the group member options.

bryanpkc:
flang-compiler/flang#1242 will fix the test failures seen in the "Build and test Flang" jobs in this PR.

clang/include/clang/Driver/Options.td
```diff
@@ -525,7 +525,7 @@ class InternalDriverOpt : Group<internal_driver_Group>,
   Flags<[NoXarchOption, HelpHidden]>;
 def driver_mode : Joined<["--"], "driver-mode=">, Group<internal_driver_Group>,
   Flags<[CoreOption, NoXarchOption, HelpHidden]>,
-  HelpText<"Set the driver mode to either 'gcc', 'g++', 'cpp', or 'cl'">;
+  HelpText<"Set the driver mode to one of: 'gcc', 'g++', 'cpp', 'cl', or 'fortran'">;
```
RichBarton-Arm: Seems like an unnecessary bit of delta in classic flang that could be removed?

clang/include/clang/Driver/Options.td
```diff
@@ -4663,37 +4679,56 @@ def ffixed_line_length_EQ : Joined<["-"], "ffixed-line-length=">, Group<f_Group>
...
+def fno_default_integer_8 : Flag<["-"], "fno-default-integer-8">, Group<f_Group>;
+#endif
 def fdefault_real_8 : Flag<["-"],"fdefault-real-8">, Group<f_Group>,
   HelpText<"Set the default real kind to an 8 byte wide type">;
+#ifdef ENABLE_CLASSIC_FLANG
+def fno_default_real_8 : Flag<["-"], "fno-default-real-8">, Group<f_Group>;
+#endif
 def flarge_sizes : Flag<["-"],"flarge-sizes">, Group<f_Group>,
   HelpText<"Use INTEGER(KIND=8) for the result type in size-related intrinsics">;
-
```
RichBarton-Arm, line 4676: Unnecessary delta

clang/include/clang/Driver/Options.td
```diff
@@ -4663,37 +4679,56 @@ def ffixed_line_length_EQ : Joined<["-"], "ffixed-line-length=">, Group<f_Group>
...
 def falternative_parameter_statement : Flag<["-"], "falternative-parameter-statement">, Group<f_Group>,
   HelpText<"Enable the old style PARAMETER statement">;
 def fintrinsic_modules_path : Separate<["-"], "fintrinsic-modules-path">,  Group<f_Group>, MetaVarName<"<dir>">,
   HelpText<"Specify where to find the compiled intrinsic modules">,
   DocBrief<[{This option specifies the location of pre-compiled intrinsic modules,
   if they are not in the default location expected by the compiler.}]>;
-
```
RichBarton-Arm, line 4683: unnecessary delta

clang/include/clang/Driver/Options.td
RichBarton-Arm: For the record, I am very happy with the new approach that lets classic flang sit next to llvm flang wrt Options.td.
Sterling work @bryanpkc

clang/include/clang/Driver/ToolChain.h
```diff
@@ -601,6 +601,14 @@ class ToolChain {
   AddClangSystemIncludeArgs(const llvm::opt::ArgList &DriverArgs,
                             llvm::opt::ArgStringList &CC1Args) const;
 
+  /// \brief Add the flang arguments for system include paths.
```
RichBarton-Arm, line 605: Needs the ENABLE_CLASSIC_FLANG guard?

clang/include/clang/Driver/ToolChain.h
```diff
@@ -693,6 +701,11 @@ class ToolChain {
   virtual llvm::SmallVector<BitCodeLibraryInfo, 12>
   getHIPDeviceLibs(const llvm::opt::ArgList &Args) const;
 
+  /// AddFortranStdlibLibArgs - Add the system specific linker arguments to use
```
RichBarton-Arm, line 707: Needs the ENABLE_CLASSIC_FLANG guard?

clang/lib/CodeGen/CGDebugInfo.cpp
```diff
@@ -3886,7 +3886,8 @@ CGDebugInfo::getGlobalVariableForwardDeclaration(const VarDecl *VD) {
   auto Align = getDeclAlignIfRequired(VD, CGM.getContext());
   auto *GV = DBuilder.createTempGlobalVariableFwdDecl(
       DContext, Name, LinkageName, Unit, Line, getOrCreateType(T, Unit),
-      !VD->isExternallyVisible(), nullptr, TemplateParameters, Align);
```
RichBarton-Arm, line 3889: Changes in file need the ENABLE_CLASSIC_FLANG guard?

clang/lib/Driver/CMakeLists.txt
```diff
@@ -82,6 +87,7 @@ add_clang_library(clangDriver
   ToolChains/ZOS.cpp
   Types.cpp
   XRayArgs.cpp
+  ${FLANG_FILES}
```
RichBarton-Arm: I wonder if that insertion might be more obvious inserted at the same line as the deletion of Flang.cpp. Minor nit though.

clang/lib/Driver/ToolChains/Clang.cpp
```diff
@@ -4956,7 +4886,68 @@ void Clang::ConstructJob(Compilation &C, const JobAction &JA,
...
+#ifdef ENABLE_CLASSIC_FLANG
+  if (Args.getLastArg(options::OPT_fveclib))
+    Args.AddLastArg(CmdArgs, options::OPT_fveclib);
+  else
+    CmdArgs.push_back("-fveclib=PGMATH");
+#else
   Args.AddLastArg(CmdArgs, options::OPT_fveclib);
+#endif
+
+  std::string PassRemarkVal(""), PassRemarkOpt("");
```
RichBarton-Arm, line 4897: I think this whole block also needs an include guard as it uses `OPT_Minfo_EQ` and friends? So I think this branch currently wouldn't build with `-DLLVM_ENABLE_CLASSIC_FLANG=Off`

RichBarton-Arm: Weird also how this block needs to be in the Clang driver and can't just be in ClassicFlang.cpp

clang/lib/Driver/ToolChains/ClassicFlang.cpp
RichBarton-Arm: I'd be very afraid to change default here. Better the devil you know.

clang/lib/Driver/ToolChains/CommonArgs.cpp
```diff
@@ -119,6 +123,23 @@ static void renderRemarksHotnessOptions(const ArgList &Args,
         Twine("--plugin-opt=opt-remarks-hotness-threshold=") + A->getValue()));
 }
 
+/// \brief Determine if Fortran "main" object is needed
```
RichBarton-Arm, line 127: Needs preprocessor guard?

RichBarton-Arm (commented):
Looking great Bryan! I like the solution to Options.td preprocessing now, thanks for taking time to do that. 

I notice that there are a number of places in the code where the classic flang changes are not guarded with `ENABLE_CLASSIC_FLANG` Some of those are harmless, but it looks to my eye like some of them need guarding after your changes to Options.td preprocessing. Worth a look before we approve.

clang/include/clang/Driver/Options.td
bryanpkc: Actually I think this change should be upstreamed, but `'fortran'` should be replaced with `'flang'`.

clang/lib/Driver/ToolChains/Clang.cpp
bryanpkc: I have guarded the whole block with `ENABLE_CLASSIC_FLANG`. I will create a follow-up issue to investigate moving this code to a more appropriate location.

clang/lib/Driver/ToolChains/CommonArgs.cpp
bryanpkc: Fixed.

clang/lib/Driver/CMakeLists.txt
bryanpkc: Fixed.

clang/include/clang/Driver/ToolChain.h
bryanpkc: Fixed.

clang/include/clang/Driver/ToolChain.h
bryanpkc: Fixed.

bryanpkc:
flang-compiler/flang#1244 is needed to fix several AArch64 Flang test cases.

clang/lib/CodeGen/CGDebugInfo.cpp
bryanpkc: Unlike the changes we made in Driver, the optional `flags` field we added to the `DIGlobalVariable` metadata has not caused much problem or confusion historically. Inserting `#ifdef` in all the places where we use the debuginfo metadata APIs will likely create much more verbose delta, but no improvement in clarity.

bryanpkc:
@RichBarton-Arm  Could you take a look at this again?

clang/lib/Driver/ToolChains/ClassicFlang.h
```diff
@@ -0,0 +1,49 @@
...
+//
+//                     The LLVM Compiler Infrastructure
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef LLVM_CLANG_LIB_DRIVER_TOOLCHAINS_Flang_H
+#define LLVM_CLANG_LIB_DRIVER_TOOLCHAINS_Flang_H
```
RichBarton-Arm: Shouldn't this be ..._ClassicFlang_H ?

RichBarton-Arm (approved):
Two comments, but from the clang/driver side I think this merge is good to go with those two addressed. It would be good for someone to also look over the Debug changes here to check they are all legit.

My two comments:
 1. It looks like this commit is not present in this port, but I expected it to be: https://github.com/flang-compiler/classic-flang-llvm-project/commit/89065d4abfe506af1da432ddf8651f1893a4a4fd @pawosm-arm  can confirm if that's right.
 2. I left one inline comment also

clang/lib/Driver/ToolChains/ClassicFlang.h
pawosm-arm: The commit mentioned by Richard should be in this port.

bryanpkc: The commit is not in this port because it is already part of the baseline llvmorg-14.0.1 (commit 26f56438e3dab44cea4c8f16d4cb16e9424b02c6).

bryanpkc: The macro has been fixed.

pawosm-arm (approved):

kiranchandramohan:
ping @shivaramaarao

shivaramaarao (approved):

PR #111, opened 2022-06-07
Fix flang driver preprocessor issue

RichBarton-Arm:
Fix a bug where flang -E -save-temps input.F90 would preprocess
the input file twice. Add a new test for preprocessor behaviour that
would expose the bug and updated fortran-phases test for new behaviour.

Also added -emit-flang-llvm to that test for completeness.

Signed-off-by: Richard Barton <richard.barton@arm.com>


pawosm-arm (approved):

RichBarton-Arm:
@bryanpkc - this is a fix for the issue that the LLVM-14 merge brings in where the preprocessor job is run twice with -save-temps. Please could you review?

bryanpkc (approved):
LGTM

PR #113, opened 2022-06-24
Changes to flang's fp-contract behaviour (release_13x)

pawosm-arm:
1) All fma and contraction behvaviour follows from the -ffp-contract= flag
setting.
2) At optimization level 0, -ffp-contract=fast flag will not be honoured.
3) At all other levels, -ffp-contract=fast will be the default behaviour and
the flag is honoured.
This commit is accompanied with a symmetric change on the classic flang repository: https://github.com/flang-compiler/flang/pull/1267

PR #118, opened 2022-06-30
Changes for reassoc attributes (release_13x)

pawosm-arm:
This patch adds code to generate the driver flags so that llvm bridge in flang2 can generate the nsz, reassoc attributes to arithmetic instructions. Also included is a testcase.

Associated commit for the flang side: https://github.com/flang-compiler/flang/pull/1268

### DIFF
--- a/clang/lib/Driver/ToolChains/ClassicFlang.cpp
+++ b/clang/lib/Driver/ToolChains/ClassicFlang.cpp
@@ -69,6 +69,8 @@ void ClassicFlang::ConstructJob(Compilation &C, const JobAction &JA,
   bool NeedIEEE = true;
   bool NeedFastMath = false;
   bool NeedRelaxedMath = false;
+  bool AssociativeMath = false;
+  bool SignedZeros = true;
 
   // Check number of inputs for sanity. We need at least one input.
   assert(Inputs.size() >= 1 && "Must have at least one input.");
@@ -507,14 +509,18 @@ void ClassicFlang::ConstructJob(Compilation &C, const JobAction &JA,
   */
   for(Arg *A: Args.filtered(options::OPT_ffast_math, options::OPT_fno_fast_math,
                         options::OPT_Ofast, options::OPT_Kieee_off,
-                        options::OPT_Kieee_on, options::OPT_frelaxed_math)) {
+                        options::OPT_Kieee_on, options::OPT_frelaxed_math,
+                        options::OPT_fassociative_math,
+                        options::OPT_fno_associative_math,
+                        options::OPT_fsigned_zeros,
+                        options::OPT_fno_signed_zeros)) {
     if (A->getOption().matches(options::OPT_ffast_math) ||
         A->getOption().matches(options::OPT_Ofast)) {
       NeedIEEE = NeedRelaxedMath = false;
       NeedFastMath = true;
     } else if (A->getOption().matches(options::OPT_Kieee_on)) {
-      NeedFastMath = NeedRelaxedMath = false;
-      NeedIEEE = true;
+      NeedFastMath = NeedRelaxedMath = AssociativeMath = false;
+      NeedIEEE = SignedZeros = true;
     } else if (A->getOption().matches(options::OPT_frelaxed_math)) {
       NeedFastMath = NeedIEEE = false;
       NeedRelaxedMath = true;
@@ -522,6 +528,16 @@ void ClassicFlang::ConstructJob(Compilation &C, const JobAction &JA,
       NeedFastMath = false;
     } else if (A->getOption().matches(options::OPT_Kieee_off)) {
       NeedIEEE = false;
+    } else if (A->getOption().matches(options::OPT_fassociative_math)) {
+      AssociativeMath = true;
+      NeedIEEE = SignedZeros = false;
+    } else if (A->getOption().matches(options::OPT_fno_associative_math)) {
+      AssociativeMath = false;
+    } else if (A->getOption().matches(options::OPT_fsigned_zeros)) {
+      SignedZeros = true;
+      AssociativeMath = false;
+    } else if (A->getOption().matches(options::OPT_fno_signed_zeros)) {
+      SignedZeros = NeedIEEE = false;
     }
     A->claim();
   }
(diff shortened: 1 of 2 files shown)


clang/test/Driver/flang/reassoc.f90
```diff
@@ -0,0 +1,19 @@
...
+! RUN: %flang -fno-trapping-math %s -### 2>&1 | FileCheck %s
+! RUN: %flang -fassociative-math -fno-trapping-math %s -### 2>&1 | FileCheck %s
+
+! RUN: %flang -fno-signed-zeros %s -### 2>&1 | FileCheck --check-prefix=NSZ %s
+! RUN: %flang -fassociative-math -fno-signed-zeros %s -### 2>&1 | FileCheck --check-prefix=NSZ %s
+! RUN: %flang -fno-signed-zeros -fno-trapping-math %s -### 2>&1 | FileCheck --check-prefix=NSZ %s
+! RUN: %flang -fassociative-math -fno-signed-zeros -fno-trapping-math %s -### 2>&1 | FileCheck --check-prefix=NSZ_REASSOCIATE %s
+
+! CHECK-NOT: {{.*}}flang2{.*}} "-x" "216" "0x8"
+! CHECK-NOT: {{.*}}flang2{.*}} "-x" "216" "0x10"
```
bryanpkc: The second regex should be `{{.*}}` on these two lines.

pawosm-arm:
The flang part was refreshed. @reviewers could you have a look?

bryanpkc (changes requested):
If `-Kieee` or `-fno-fast-math` is specified after `-fassociative-math` or `-fno-signed-zeros`, they should override the earlier flags. IIUC, these related options should be handled together.

bryanpkc:
The same comment applies to the `fp-contract` patches (#113, etc.) too. `-ffp-contract=` should be overridden by `-Kieee` and `-fno-fast-math`.

pawosm-arm:
> The same comment applies to the `fp-contract` patches (#113, etc.) too. `-ffp-contract=` should be overridden by `-Kieee` and `-fno-fast-math`.

I'm not sure about this one. AFAIR `fp-contract=fast` is about emitting FMA's, it's being 'fast' is not at cost of accuracy. Fast-math is not the same 'fast'.

pawosm-arm:
> > The same comment applies to the `fp-contract` patches (#113, etc.) too. `-ffp-contract=` should be overridden by `-Kieee` and `-fno-fast-math`.
> 
> I'm not sure about this one. AFAIR `fp-contract=fast` is about emitting FMA's, it's being 'fast' is not at cost of accuracy. Fast-math is not the same 'fast'.

I've looked into the Clang sources, and I can't see how -ffp-contract could be overriden by force of demanding IEEE or no-fast-math.
The default is to use fp contraction, unless specifically denied (e.g. by specifying fp-model, which flang doesn't consider anyway), then this could be overriden in Clang as such:
```
    case options::OPT_ffast_math:
...
      // If fast-math is set then set the fp-contract mode to fast.
      FPContract = "fast";
```
So 'fast' fp contraction is being in favor. Following this flag definition: "-ffp-contract=fast enables floating-point expression contraction such as forming of fused multiply-add (FMA) operations if the target has native support for them.", "FMA operations can speed up and improve the accuracy of many computations that involve the accumulation of products". There's nothing about accuracy degradation or violating IEEE.
Clang does a lot around FpContract, but it's mostly because of `-ffp-model` flag which flang doesn't consider anyway.
From this point of view, what classic flang frontend driver currently does (for the fp contraction) is pretty much compatible with what clang does and does not need to be changed.

bryanpkc:
> I've looked into the Clang sources, and I can't see how -ffp-contract could be overriden by force of demanding IEEE or no-fast-math.

If I understand https://github.com/flang-compiler/classic-flang-llvm-project/blob/release_14x/clang/lib/Driver/ToolChains/Clang.cpp#L2890-L2897 correctly, `-fno-fast-math` will override `-ffp-contract=fast` and change the `fp-contract` setting to `on`.

pawosm-arm:
> > I've looked into the Clang sources, and I can't see how -ffp-contract could be overriden by force of demanding IEEE or no-fast-math.
> 
> If I understand https://github.com/flang-compiler/classic-flang-llvm-project/blob/release_14x/clang/lib/Driver/ToolChains/Clang.cpp#L2890-L2897 correctly, `-fno-fast-math` will override `-ffp-contract=fast` and change the `fp-contract` setting to `on`.

That's true, but we didn't differentiate between fast and on, the EnableFPContraction variable is just a simple boolean stating that fp contraction is or isn't happening, hence Clang's logic couldn't be fully replicated.

bryanpkc:
> hence Clang's logic couldn't be fully replicated.

Ah okay, you're right about `-ffp-contract=` then. Thanks for your consideration.

pawosm-arm:
I have something like this waiting in a pipeline (checking for `NeedIEEE` when setting xflags, along with changes in the test case):
```
+  if (!SignedZeros && !NeedIEEE) {
+    LowerCmdArgs.push_back("-x");
+    LowerCmdArgs.push_back("216");
+    LowerCmdArgs.push_back("0x8");
+  }
+  if (AssociativeMath && !SignedZeros && !NeedIEEE) {
+    LowerCmdArgs.push_back("-x");
+    LowerCmdArgs.push_back("216");
+    LowerCmdArgs.push_back("0x10");
+  }
```
sadly, github has stuck on 'push', so I'll try again tomorrow.

pawosm-arm:
Hello @reviewers, can I ask you to look at it again?

bryanpkc:
@pawosm-arm I don't think notifying "reviewers" does what you think it does. 😄

clang/test/Driver/flang/reassoc.f90
```diff
@@ -0,0 +1,19 @@
+! REQUIRES: classic_flang
+
+! Tests for flags which generate nsw, reassoc attributes
+
+! RUN: %flang -Kieee %s -### 2>&1 | FileCheck %s
+! RUN: %flang -Kieee -fno-signed-zeros %s -### 2>&1 | FileCheck %s
+! RUN: %flang -Kieee -fassociative-math %s -### 2>&1 | FileCheck %s
+! RUN: %flang -Kieee -fassociative-math -fno-signed-zeros %s -### 2>&1 | FileCheck %s
+! RUN: %flang -Knoieee %s -### 2>&1 | FileCheck %s
+! RUN: %flang -Knoieee -fassociative-math %s -### 2>&1 | FileCheck %s
```
bryanpkc: Add coverage for other options that turn off `NeedIEEE`, e.g. `-ffast-math`, `-frelaxed-math`.

clang/lib/Driver/ToolChains/ClassicFlang.cpp
```diff
@@ -1006,6 +1006,28 @@ void ClassicFlang::ConstructJob(Compilation &C, const JobAction &JA,
...
+    case options::OPT_fno_signed_zeros:     SignedZeros = false;      break;
+    default: break;
+    }
+  }
+  if (!SignedZeros && !NeedIEEE) {
+    LowerCmdArgs.push_back("-x");
+    LowerCmdArgs.push_back("216");
+    LowerCmdArgs.push_back("0x8");
+  }
+  if (AssociativeMath && !SignedZeros && !NeedIEEE) {
```
bryanpkc: This implementation allows an earlier option to suppress the effect of a subsequent option, which contradicts the usual understanding of how compiler options are processed. I would prefer that Classic Flang processes options the same way as Clang would; if it doesn't, then I think we should clearly document the difference in behaviour. @shivaramaarao What do you think?

clang/test/Driver/flang/reassoc.f90
pawosm-arm: added

pawosm-arm:
@bryanpkc could you look at the latest changes?

bryanpkc (approved):
LGTM. Thanks for reworking the patch!

PR #120, opened 2022-06-30
Changes for reassoc attributes (release_11x)

pawosm-arm:
This patch adds code to generate the driver flags so that llvm bridge in
flang2 can generate the nsz, reassoc attributes to arithmetic instructions.
Also included is a testcase.

Associated commit for the flang side: https://github.com/flang-compiler/flang/pull/1268


pawosm-arm:
we don't test release_11x anymore.

PR #122, opened 2022-07-15
classic flang: enable IEEE by default (and sort Lower/Common inconsistency out)

pawosm-arm:


pawosm-arm:
There is a symmetric patch with needs to be applied on the flang side in order for this to pass: https://github.com/flang-compiler/flang/pull/1276

shivaramaarao (approved):

clang/lib/Driver/ToolChains/ClassicFlang.cpp
```diff
@@ -585,8 +585,8 @@ void ClassicFlang::ConstructJob(Compilation &C, const JobAction &JA,
     LowerCmdArgs.push_back("216");
     LowerCmdArgs.push_back("1");
     // Lower: -ieee 0
-    CommonCmdArgs.push_back("-ieee");
-    CommonCmdArgs.push_back("0");
+    LowerCmdArgs.push_back("-ieee");
+    LowerCmdArgs.push_back("0");
```
bryanpkc: Should we not pass `-ieee 0` to `flang1`?

bryanpkc (approved):
LGTM

PR #123, opened 2022-07-15
classic flang: enable IEEE by default (and sort Lower/Common inconsistency out)

pawosm-arm:


pawosm-arm:
There is a symmetric patch with needs to be applied on the flang side in order for this to pass: flang-compiler/flang#1276

PR #126, opened 2022-07-16
[workflows] Fix CI for release_14x

bryanpkc:
1. Parameterize `Ubuntu20Dockerfile` with the branch name. Previously the Dockerfile was still creating container images from the hardcoded release_13x branch.

2. Update the workaround for MacOS X build failures according to https://github.com/actions/virtual-environments/issues/2274, to avoid the "no member named 'xxx' in the global namespace" problems when including `<cmath>`.

3. For debugging purpose, add a -v option to build-llvm-project.sh to enable verbose output. This option sets `$VERBOSE` to a non-empty string which forces the CMake-generated Makefile to print every command it runs.
 
4. Delete the repo-lockdown workflow to eliminate the job which is always a no-op for classic-flang-llvm-project.


pawosm-arm (approved):

bryanpkc:
@shivaramaarao Would you like to review/approve this PR? I'll merge it if I don't hear from you within a day, since it's a CI fix.

shivaramaarao (approved):

PR #127, opened 2022-07-28
[build] Stop caching AArch64 Docker image layers

bryanpkc:
The "Pre-compile llvm ARM64" workflow tries to cache Docker image layers to avoid rebuilding the same content over and over, as PRs are merged into the release branches. Mysteriously, in older release branches the workflow was not successfully caching layers, while in release_14x, it is too successful, resulting in the same image being uploaded for successive commit hashes. The actual build process is not too time-consuming (about 15 minutes) so it is safer and simpler to turn off caching completely.

### DIFF
--- a/.github/workflows/build_push_docker_image_Ubuntu20.yml
+++ b/.github/workflows/build_push_docker_image_Ubuntu20.yml
@@ -22,14 +22,6 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
 
-      - name: Cache Docker layers
-        uses: actions/cache@v2
-        with:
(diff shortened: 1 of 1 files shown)


bryanpkc:
This is an example of a `release_14x` job which "caches" the layer that is supposed to build LLVM:

https://github.com/flang-compiler/classic-flang-llvm-project/runs/7469978551#step:7:150

And this is an example of a `release_13x` job which actually builds LLVM from source and creates a new layer:

https://github.com/flang-compiler/classic-flang-llvm-project/runs/7469966379#step:7:210

pawosm-arm (approved):

PR #130, opened 2022-08-01
[llvm] Provide separate veclib mapping of PGMATH functions for AArch64 (release_14x)

pawosm-arm:
This patch fixes link-time issues experienced while compiling some
more obscure Fortran workloads. For those programs, calls to the
libpgmath functions not defined for AArch64 were generated, e.g.
`__pg_log_4` or `__fd_log_4`. This patch eliminates such possibility.

### DIFF
--- a/clang/test/CodeGen/libpgmath-logfun-aarch64.ll
+++ b/clang/test/CodeGen/libpgmath-logfun-aarch64.ll
@@ -1,0 +1,60 @@
+; REQUIRES: aarch64-registered-target
+
+; RUN: %clang -target aarch64-unknown-linux-gnu -Ofast -S %s -o - | FileCheck %s
+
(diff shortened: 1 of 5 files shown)


bryanpkc:
We have done something similar downstream, but for `release_14x` branch, we moved these libpgmath function mappings into `llvm/include/llvm/Analysis/VecFuncs.def`, as is done for all other libraries handled by TargetLibraryInfo.cpp.

bryanpkc (changes requested):
Please refactor this patch to use `llvm/Analysis/VecFuncs.def` as described in my previous comment.

pawosm-arm:
The macOS bug is unrelated to this commit.

pawosm-arm:
> Please refactor this patch to use `llvm/Analysis/VecFuncs.def` as described in my previous comment.

@bryanpkc is it ok now?

bryanpkc (approved):
Yep, LGTM. Thanks for refactoring!

PR #131, opened 2022-08-08
[Driver] Remove the "-lompstub" in the driver (release_14x)

PeixinQiao:
In some workloads, users get "-lompstub" by running "flang -v" and link
it explicitly, which may cause failure when compiling programs with
OpenMP (flang -fopenmp test.f90 -lompstub).

There is no runtime call in ompstub.c generated when "-fopenmp" is not
added to compile the program. When "-fopenmp" is added to compile the
program, the runtime in llvm/openmp(libomp.so) will be used. If users
use some runtime library routines such as "omp_get_thread_num" and
compiles it without the option "-fopenmp", both gfortran and ifort
report errors of "undefined reference to ...". After remove "-lompstub"
in the driver, classic-flang reports the error, too. So, it's safe to
remove the "-lompstub" in the driver.

### DIFF
--- a/clang/lib/Driver/ToolChain.cpp
+++ b/clang/lib/Driver/ToolChain.cpp
@@ -1019,8 +1019,6 @@ void ToolChain::AddFortranStdlibLibArgs(const ArgList &Args,
   CmdArgs.push_back("-lpgmath");
   if (useOpenMP)
     CmdArgs.push_back("-lomp");
-  else
-    CmdArgs.push_back("-lompstub");
   if (staticFlangLibs)
     CmdArgs.push_back("-Bdynamic");
 


PeixinQiao:
ping @bryanpkc @pawosm-arm @kiranchandramohan @shivaramaarao 
I cannot add reviewers in this project.

bryanpkc (approved):

pawosm-arm:
ompstubs were our headache some time ago and we still not decided to get rid of them. The lack of openmp symbols was causing link time troubles on edge cases, sadly I don't remember the details now. Did you analyze why PGI decided to add them?

PeixinQiao:
The ompstubs was added in the first commit of classic flang without explanations. The only thing I can find is "[libpgc stubs for omp_ routines not appearing in crit.o & omp.o for non-x64/x86 targets](https://github.com/flang-compiler/flang/blob/0a63d55c82a548677b82095e48cb2043ee182db4/runtime/ompstub/ompstubs.c#L9-L12)".

bryanpkc:
@shivaramaarao @pawosm-arm Does this PR look problematic to you? If not, I'd like to merge this.

shivaramaarao:
> 

couple of testcases seem to fail 
  f90_correct/src/rio2.f90 and f90_correct/src/rio3.f90

both of these tests seem to have calls to omp runtime library calls and are compiled without -fopenmp flag.

bryanpkc:
@shivaramaarao Please make sure you have picked up https://github.com/flang-compiler/flang/pull/1282.

shivaramaarao:
> @shivaramaarao Please make sure you have picked up [flang-compiler/flang#1282](https://github.com/flang-compiler/flang/pull/1282).

ok. my bad. this is already addressed. LGTM

shivaramaarao (approved):

pawosm-arm:
> @shivaramaarao @pawosm-arm Does this PR look problematic to you? If not, I'd like to merge this.

@bryanpkc the whole OpenMP stubs issue is problematic to me as I don't really remember why we had addressed the issue differently, e.g. all of the symbols in our libompstubs are weak so they could be hidden behind the real OpenMP runtime functions when needed (AFAIR, PGI did not like this idea either, so we're having it only downstream). I'm trying to recall what real life problem occurred when we tried to get rid of libompstubs, but nothing really comes to my mind.

bryanpkc:
`init_nomp.c` defines a number of `_mp_*` functions which are called by generated code directly; see for examples:

bryanpkc:
`runtime/ompstub/init_nomp.c` defines a number of `_mp_*` stub functions which are called by generated code directly; see for examples:

- https://github.com/flang-compiler/flang/blob/master/tools/flang2/flang2exe/exp_ftn.cpp#L2067-L2089
- https://github.com/flang-compiler/flang/blob/master/tools/flang2/utils/ilmtp/x86_64/ilmtp.n#L2857-L2890

They are probably only emitted when `-fopenmp` is used, but I'm not 100% sure. They are not OpenMP APIs and so technically shouldn't be in libompstub; perhaps they should be moved into `runtime/flangrti/` like many other `_mp_*` functions.

It is also possible that these `_mp_*` stub functions in libompstub are either never used, or are used incorrectly in multi-threaded programs, because I can't find any "non-stub" implementation of the same functions anywhere.

PeixinQiao:
The following functions only exist in `init_nomp.c`:
```
$ grep -rn "_mp_avlcpus" .
./runtime/ompstub/init_nomp.c:22:_mp_avlcpus()
$ grep -rn "_mp_get_tcpus" .
./runtime/ompstub/init_nomp.c:39:_mp_get_tcpus()
./runtime/ompstub/init_nomp.c:94:_mp_get_tcpus_max()
$ grep -rn "_mp_get_par" .
./runtime/ompstub/init_nomp.c:102:_mp_get_par()
./runtime/ompstub/init_nomp.c:110:_mp_get_parpar(int n)
$ grep -rn "_mp_pcpu_team" .
./runtime/ompstub/init_nomp.c:118:_mp_pcpu_team(void)
$ grep -rn "_mp_get_pteam" .
./runtime/ompstub/init_nomp.c:124:_mp_get_pteam(void * t) 
```
The `_mp_lcpu*` and `_mp_ncpu*` may be generated elsewhere. According to the desciption in `ilmtp.n`, `_mp_lcpu2` is generated for `sections` directive, but I tested and no such function generated in IR. But to be safe, I move `init_nomp.c` to flangrti to follow `llcrit.c` https://github.com/flang-compiler/flang/pull/1283.

PeixinQiao:
@pawosm-arm Can this PR be approved now?

pawosm-arm (approved):

PR #134, opened 2022-08-29
Add platform independent build script for LLVM (release_14x)

tdusnoki:


scripts/build_llvm_project.py
```diff
@@ -0,0 +1,139 @@
...
+                        help='install directory (default: %(default)s) ')
+  buildopt.add_argument('-j', '--jobs', metavar='N', type=int, default=os.cpu_count(),
+                        help='number of parallel build jobs (default: %(default)s)')
+  buildopt.add_argument('--toolchain', metavar='FILE', default=default_toolchain(),
+                        help='specify toolchain file (default: %(default)s)')
+  buildopt.add_argument('--builddir', metavar='DIR', default='build',
+                        help=f'specify build directory (default: {settings.LLVM_DIR}/%(default)s)')
+  buildopt.add_argument('--clean', action='store_true', default=False,
+                        help='clean build')
+  buildopt.add_argument('--build-type', metavar='TYPE', default='Release',
```
bryanpkc: I think the options should support a single-letter short form, similar to the ones in `build-llvm-project.sh`. `--build-type` and `--use-ccache` are especially important since a user will probably be typing them quite frequently.

scripts/build_llvm_project.py
```diff
@@ -0,0 +1,139 @@
...
+                        help='set build type (default: %(default)s)')
+  buildopt.add_argument('--debug', dest='build_type', action='store_const', const='Debug', default=argparse.SUPPRESS,
+                        help='debug build (alias for --build-type %(const)s)')
```
bryanpkc: As I suggested in https://github.com/flang-compiler/flang/pull/1185#discussion_r989001246, please remove `--debug` (since you already have `--build-type`) and add `--verbose`.

scripts/build_llvm_project.py
```diff
@@ -0,0 +1,151 @@
...
+  buildopt = parser.add_argument_group('general build options')
+  buildopt.add_argument('-t', '--target', metavar='ARCH', choices=['X86', 'AArch64', 'PowerPC'], default='X86',
+                        help='Control which targets are enabled (%(choices)s) (default: %(default)s)')
+  buildopt.add_argument('-p', '--install-prefix', metavar='PATH', nargs='?', default=Path(settings.LLVM_DIR,'..','install'), const=False,
+                        help='install directory (default: %(default)s) ')
+  buildopt.add_argument('-j', '--jobs', metavar='N', type=int, default=os.cpu_count(),
+                        help='number of parallel build jobs (default: %(default)s)')
+  buildopt.add_argument('--toolchain', metavar='FILE', default=default_toolchain().as_posix(),
+                        help='specify toolchain file (default: %(default)s)')
+  buildopt.add_argument('--builddir', metavar='DIR', default='build',
```
bryanpkc: Could you make `-d` an alias of this option, to make this script more consistent with `flang/scripts/build_flang.py`?

scripts/build_llvm_project.py
```diff
@@ -0,0 +1,151 @@
...
+                        help=f'specify build directory (default: {settings.LLVM_DIR}/%(default)s)')
+  buildopt.add_argument('--clean', action='store_true', default=False,
+                        help='clean build')
+  buildopt.add_argument('-d', '--build-type', metavar='TYPE', default='Release',
```
bryanpkc: Could you make `-b` an alias of this option, to make this script more consistent with `flang/scripts/build_flang.py`?

scripts/build_llvm_project.py
```diff
@@ -0,0 +1,151 @@
...
+                        help='set build type (default: %(default)s)')
+  buildopt.add_argument('-x', '--cmake-param', metavar='OPT', action='append', default=[],
+                        help='add custom argument to CMake')
+  buildopt.add_argument('-e', '--llvm-enable-projects', metavar='OPT', action='append', default=['clang'] if sys.platform == 'win32' else ['clang;openmp'],
+                        help='enable llvm projects to build (in quotation marks separated by semicolons e.g.: "clang;openmp")')
+  buildopt.add_argument('-c', '--use-ccache', action="store_true", default=False,
+                        help='Using ccache during the build (default: %(default)s)')
+  buildopt.add_argument('-a', '--c-compiler', metavar='OPT', default=None,
```
bryanpkc: `-a` was never a good mnemonic for this option. I also don't think this will be used very often since we can use `--toolchain`. I would suggest renaming this to `--cc`, without a mnemonic.

scripts/build_llvm_project.py
```diff
@@ -0,0 +1,151 @@
...
+                        help='use specific C compiler')
+  buildopt.add_argument('-b', '--cxx-compiler', metavar='OPT', default=None,
```
bryanpkc: Since `-b` should be an alias of `--build-type`, could you rename this to just `--cxx`, without a mnemonic?

bryanpkc (changes requested):

bryanpkc (approved):

kaadam (approved):
LGTM

PR #136, opened 2022-10-20
Delete pgmath veclib definitions for sincos for release14

gbaraldi:
Rebase of #132 on release_14x. I also deleted the aarch64 functions, but I haven't tested if they caused problems like the x86 ones. A rough guess would be that they also do, because pgmath does that weird vector thing.

@Keno

### DIFF
--- a/llvm/include/llvm/Analysis/VecFuncs.def
+++ b/llvm/include/llvm/Analysis/VecFuncs.def
@@ -487,6 +487,15 @@ TLI_DEFINE_VECFUNC("__ps_cos_1", "__ps_cos_4", FIXED(4))
 TLI_DEFINE_VECFUNC("__rd_cos_1", "__rd_cos_2", FIXED(2))
 TLI_DEFINE_VECFUNC("__rs_cos_1", "__rs_cos_4", FIXED(4))
 
+TLI_DEFINE_VECFUNC("__fd_sincos_1", "__fd_sincos_2", FIXED(2))
+TLI_DEFINE_VECFUNC("__fs_sincos_1", "__fs_sincos_4", FIXED(4))
+
+TLI_DEFINE_VECFUNC("__pd_sincos_1", "__pd_sincos_2", FIXED(2))
+TLI_DEFINE_VECFUNC("__ps_sincos_1", "__ps_sincos_4", FIXED(4))
+
+TLI_DEFINE_VECFUNC("__rd_sincos_1", "__rd_sincos_2", FIXED(2))
+TLI_DEFINE_VECFUNC("__rs_sincos_1", "__rs_sincos_4", FIXED(4))
+
 TLI_DEFINE_VECFUNC("__fd_tan_1", "__fd_tan_2", FIXED(2))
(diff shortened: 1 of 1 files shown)


pawosm-arm (changes requested):
There are two issues here, a small one and a big one.
A small one: the commit message should reflect what the commit actually does (apart from being merely rebase)
A big one: the 'I haven't tested' part of PR message is a red flag. This commit lacks a test case: something that would fail without this commit and passes with it. I.e. it would make it so much easier for me to e.g. verify that the AArch64 change is actually needed.

gbaraldi:
@pawosm-arm 
From #11 it seems that
```fortran
program foo
      real(kind=8) :: val, res1, res2
      read *, val
      res1 = sin(val)
      res2 = cos(val)
      print *, res1, res2
end program foo
```
This fails to build with O1 before this PR, but works with it. Where should this test go?

pawosm-arm:
I've forgotten we're not in the (classic) flang repo. This test would have to go to f90_correct directory in the other repo, which would create artificial detachment we'd like to avoid.

I've tried your reproducer on AArch64, and it didn't fail. In the end, my commit that you're modifying here appeared after my failed attempt to build SPEC2k17. The things that were missing for AArch64 were already removed.

I'll try to build libpgmath for AArch64 a different way, I'll let you know if it still fails to fail.

I'm testing debug build with assertsions turned on.

pawosm-arm:
No matter how I try, it (the reproducer) isn't failing on AArch64.

Could you prepare a PR on classic flag repo with your reproducer in the f90_correct directory? In theory, it should fail on x86_64 and pass on AArch64. Depending on the outcome, we shall approve this PR or not.

pawosm-arm:
The reproducer cannot still prove that the problem occurs on AArch64, even after we have added assertions-enabled builds in our CI. Therefore I would like to ask the contributor to remove the AArch64 part from the commit.

bryanpkc:
I have just restarted the CIs on flang-compiler/flang#1310 with assertions-enabled builds. The last time they were run (27 days ago), assertions had not been enabled yet.

pawosm-arm:
> I have just restarted the CIs on [flang-compiler/flang#1310](https://github.com/flang-compiler/flang/pull/1310) with assertions-enabled builds. The last time they were run (27 days ago), assertions had not been enabled yet.

Yeah, but the failure on x86 is what we expected, while AArch64 failure is about opaque pointers, which is rather odd... Did we enable opaque pointers for release_14x branch?

bryanpkc:
@pawosm-arm 

>　Did we enable opaque pointers for release_14x branch?

No, opaque pointers are not enabled for the `release_14x` branch. However, https://github.com/flang-compiler/flang/pull/1310 needs to be rebased, and the PR itself edited, to target `legacy` instead of `master`. Since last week, the `master` branch generates opaque pointer IR and expects to work with `release_15x` branch only.

pawosm-arm (approved):
Merge instruction: please squash the commits while merging.
And cherry-pick to release_15x branch please.

bryanpkc (approved):

shivaramaarao (approved):

PR #137, opened 2022-11-02
[driver] Allow -msve-vector-bits=<n>+ syntax to mean no maximum vscale (release_14x)

LittleMeepo:
Make Flang's behaviour consistent with Clang's.

### DIFF
--- a/clang/lib/Driver/ToolChains/ClassicFlang.cpp
+++ b/clang/lib/Driver/ToolChains/ClassicFlang.cpp
@@ -1054,37 +1054,50 @@ void ClassicFlang::ConstructJob(Compilation &C, const JobAction &JA,
   }
 
   // Add vscale range
-  unsigned vscaleMin = 1U;
-  unsigned vscaleMax = 16U;
-  bool hasVscaleRange = false;
+  unsigned vscaleMin = 0;
+  unsigned vscaleMax = 0;
+  bool hasSVE = false;
   if (Arg *A = Args.getLastArg(options::OPT_msve_vector_bits_EQ)) {
     StringRef Val = A->getValue();
-
-    if (Val.equals("scalable"))
-      hasVscaleRange = true;
-    else {
-      unsigned bits = (std::stoul(Val.str()) >> 7U);
-
-      if ((bits < vscaleMin) || (bits > vscaleMax)) {
-        D.Diag(diag::warn_drv_clang_unsupported) << A->getAsString(Args);
-        hasVscaleRange = false;
-      } else {
-        vscaleMin = vscaleMax = bits;
-        hasVscaleRange = true;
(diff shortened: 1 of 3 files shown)


LittleMeepo:
The purpose of this patch is to behave in the same way as Clang.

LittleMeepo:
Corresponding PR in flang
https://github.com/flang-compiler/flang/pull/1319

pawosm-arm (changes requested):
It's good to reflect clang's behavior, but this commit is missing test cases supporting it...

LittleMeepo:
> It's good to reflect clang's behavior, but this commit is missing test cases supporting it...

Done.

clang/lib/Driver/ToolChains/ClassicFlang.cpp
```diff
@@ -1054,37 +1054,50 @@ void ClassicFlang::ConstructJob(Compilation &C, const JobAction &JA,
...
+    if (Val.equals("128") || Val.equals("256") || Val.equals("512") ||
+        Val.equals("1024") || Val.equals("2048") || Val.equals("128+") ||
+        Val.equals("256+") || Val.equals("512+") || Val.equals("1024+") ||
+        Val.equals("2048+")) {
+      unsigned Bits = 0;
+      if (Val.endswith("+"))
+        Val = Val.substr(0, Val.size() - 1);
+      else {
+        bool Invalid = Val.getAsInteger(10, Bits); (void)Invalid;
+        assert(!Invalid && "Failed to parse value");
```
bryanpkc, line 1071: Why not put the function call inside the `assert`? e.g.
```c++
assert(!Val.getAsInteger(10, Bits) && "invalid argument for -msve_vector_bits");
```

LittleMeepo: That's OK. But I just keep the format in Clang. Line 1060~1081 is copied from Clang.cpp

bryanpkc: OK.

bryanpkc (approved):
LGTM

bryanpkc:
@pawosm-arm Could you take another look?

pawosm-arm (approved):

bryanpkc:
@shivaramaarao This is an AArch64-specific patch. Would you still like to take a look?

LittleMeepo:
I realized Flang use `-vscale_range_min` and `-vscale_range_max` as parameters to parse argument, so I changed all the `-vscale-range-min` and `-vscale-range-max` to `-vscale_range_min` and `-vscale_range_max`.

bryanpkc:
@LittleMeepo Can you submit a PR with the same change for the `release_15x` branch?

PR #145, opened 2022-11-30
Add platform independent build script for LLVM (release_15x)

tdusnoki:


bryanpkc (changes requested):
My comments for #134 apply here as well.

bryanpkc (approved):
LGTM

scripts/build_llvm_project.py
```diff
@@ -0,0 +1,151 @@
...
+  build_path = normalize_builddir(
+    settings.LLVM_DIR, arguments.builddir, arguments.clean)
+
+  build_options = generate_buildoptions(arguments)
+  additional_options = [
+    f'-DLLVM_TARGETS_TO_BUILD={arguments.target}',
+    '-DLLVM_ENABLE_CLASSIC_FLANG=ON',
+  ]
+  if(arguments.llvm_enable_projects):
+    additional_options.append('-DLLVM_ENABLE_PROJECTS='+'; '.join(['%s']*len(arguments.llvm_enable_projects)) % tuple(arguments.llvm_enable_projects))
```
kaadam: Seems this expression is a bit complex, I think we can simplify it a bit.
What about this?

`'-DLLVM_ENABLE_PROJECTS={}'.format(";".join([str(item) for item in arguments.llvm_enable_projects]))
`
Maybe you can leave the generator expression so just use 
`
'-DLLVM_ENABLE_PROJECTS={}'.format(";".join(arguments.llvm_enable_projects))
`
Of course you should change the default value in the argument definition -> default=['clang', 'openmp']

kaadam (approved):
LGTM

PR #147, opened 2023-01-24
win: Don't install llvm by default (release_15x)

kaadam:
The 'llvm_build_project' build script installs llvm after the build by default. It is not necessary, if the user doesn't want.


bryanpkc (approved):

bryanpkc:
Merging CI-related changes with one approval.

PR #149, opened 2023-01-24
Setup Windows x64/arm64 GitHub action for Classic Flang LLVM

kaadam:
* Added a GitHub hosted Windows action to produce proper binaries for Flang.
* Enable running tests on Windows x64
* Updated install Python step, since CMake chooses the most recent version of 32bit Python libraries, which is 3.11 on 'windows-latest' GitHub image. Using 32bit version of Python causes a 'memory error' during the tests.

```
[856/859] Building RC object unittests\tools\llvm-mca\CMake-- Testing: 58046 tests, 2 workers --
Exception in thread Thread-13 (_handle_results):
Traceback (most recent call last):
  File "C:\hostedtoolcache\windows\Python\3.11.1\x86\Lib\threading.py", line 1038, in _bootstrap_inner
    self.run()
  File "C:\hostedtoolcache\windows\Python\3.11.1\x86\Lib\threading.py", line 975, in run
    self._target(*self._args, **self._kwargs)
  File "C:\hostedtoolcache\windows\Python\3.11.1\x86\Lib\multiprocessing\pool.py", line 579, in _handle_results
    task = get()
           ^^^^^
  File "C:\hostedtoolcache\windows\Python\3.11.1\x86\Lib\multiprocessing\connection.py", line 249, in recv
    buf = self._recv_bytes()
          ^^^^^^^^^^^^^^^^^^
  File "C:\hostedtoolcache\windows\Python\3.11.1\x86\Lib\multiprocessing\connection.py", line 317, in _recv_bytes
    return self._get_more_data(ov, maxsize)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\hostedtoolcache\windows\Python\3.11.1\x86\Lib\multiprocessing\connection.py", line 339, in _get_more_data
    ov, err = _winapi.ReadFile(self._handle, left, overlapped=True)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
MemoryError
```

### DIFF
--- a/.github/workflows/llvm-project-tests.yml
+++ b/.github/workflows/llvm-project-tests.yml
@@ -40,29 +40,49 @@ jobs:
           # We're using a specific version of macOS due to:
           # https://github.com/actions/virtual-environments/issues/5900
           - macOS-11
+          - windows-latest
+        include:
+          # Enable Windows on ARM build, when an official
+          # self-hosted machine is available.
+          # - os: self-hosted
+          #   target: AArch64
+          #   arch: arm64
+          - os: windows-latest
+            arch: amd64
+
     steps:
       - name: Setup Windows
-        if: startsWith(matrix.os, 'windows')
+        if: startsWith(matrix.os, 'windows') || startsWith(matrix.os, 'self')
         uses: llvm/actions/setup-windows@main
         with:
-          arch: amd64
-      # On Windows, starting with win19/20220814.1, cmake choose the 32-bit
-      # python3.10.6 libraries instead of the 64-bit libraries when building
-      # lldb.  Using this setup-python action to make 3.10 the default
-      # python fixes this.
+          arch: ${{ matrix.arch }}
+      # By default CMake chooses the most recent 32-bit libraries instead of
+      # the 64-bit libraries when building and testing LLVM or Clang.
+      # 32 bit version causes a memory error during the testing.
+      # On 'windows-latest' GitHub image (VS2022) 3.11 (32bit) version is installed.
+      # Using this setup-python action to make 3.11 (64bit) the default
       - name: Setup Python
+        if: ${{ matrix.os != 'self-hosted' }}
         uses: actions/setup-python@v4
         with:
-          python-version: '3.10'
+          python-version: '3.11'
       - name: Install Ninja
         uses: llvm/actions/install-ninja@main
       # actions/checkout deletes any existing files in the new git directory,
       # so this needs to either run before ccache-action or it has to use
       # clean: false.
+      - name: Setup psutil
+        # psutil is not available on Windows on ARM.
+        if: startsWith(matrix.os, 'windows')
+        run: pip install psutil
+      - name: Check tools
+        run: python --version
       - uses: actions/checkout@v3
         with:
           fetch-depth: 250
       - name: Setup ccache
+        # ccache is not available for Windows on ARM.
+        if: ${{ matrix.os != 'self-hosted' }}
         uses: hendrikmuhs/ccache-action@v1.2
         with:
           # A full build of llvm, clang, lld, and lldb takes about 250MB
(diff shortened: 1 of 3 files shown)


.github/workflows/pre-compile_llvm_windows.yml
```diff
@@ -0,0 +1,49 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup Windows
+        uses: llvm/actions/setup-windows@main
+        with:
+          arch: amd64
+
+      - name: Setup Ninja
+        run: choco install ninja
```
bryanpkc: `.github/workflows/llvm-project-tests.yml` uses the following; consider sticking to it:
```yaml
uses: llvm/actions/install-ninja@main
```

bryanpkc:
Why doesn't this job run any test?

kaadam:
> Why doesn't this job run any test?

Yes, that's a good point. I haven't included yet. Running llvm+clang tests would be enough? Shall I use existing action for that?

bryanpkc:
> Yes, that's a good point. I haven't included yet. Running llvm+clang tests would be enough? Shall I use existing action for that?

My suggestion is to add your current code into `pre-compile_llvm.yml`, which is also designed to run after a successful merge and create binaries that have passed PR reviews and CI tests. For testing PRs, I suggest you add Windows to the matrix in `llvm-project-tests.yml`.

.github/workflows/llvm-project-tests.yml
```diff
@@ -50,10 +51,12 @@ jobs:
       # python3.10.6 libraries instead of the 64-bit libraries when building
       # lldb.  Using this setup-python action to make 3.10 the default
       # python fixes this.
+      # CMake choose the most recent version of Python (32bit) which installed
+      # in 'windows-latest' (VS2022) GitHub image.
```
bryanpkc: The newly added code and comment now conflict with the original comment above. Please merge and rewrite both comments.

bryanpkc:
Have you tested the `build-win64` job in `pre-compile_llvm.yml` in your fork? If you have, can you paste the link to your job output? Since this job runs post-merge, it is not possible to test its configuration when reviewing this pull request.

kaadam:
> Have you tested the `build-win64` job in `pre-compile_llvm.yml` in your fork? If you have, can you paste the link to your job output? Since this job runs post-merge, it is not possible to test its configuration when reviewing this pull request.

Bryan, yes, sorry I forgot to place the link here:
This was the latest run: https://github.com/kaadam/classic-flang-llvm-project/actions/runs/4088129300

.github/workflows/llvm-project-tests.yml
kaadam: Okay, I'm updating the comment.

kaadam:
@bryanpkc I'm thinking to add a new Windows x64 action for creating prebuilt LLVM to Windows on ARM. However we cannot run tests on it. What do you think? Is that acceptable if I add an only cross-build job?

bryanpkc:
@kaadam Is the Windows on ARM build always cross-compiled? Can we build Flang natively on Windows on ARM? If so, it would be possible to create a self-hosted runner for the latter.

If that's difficult, I don't have a problem with the cross-compilation job. Maybe in a separate PR though?

bryanpkc (approved):

kaadam:
> @kaadam Is the Windows on ARM build always cross-compiled? Can we build Flang natively on Windows on ARM? If so, it would be possible to create a self-hosted runner for the latter.
> 
> If that's difficult, I don't have a problem with the cross-compilation job. Maybe in a separate PR though?

@bryanpkc  Basically Flang build is running on Windows on ARM natively, we only cross-compile classic-llvm for Windows on ARM, since the build was faster on a x64 host machine. Now we have a proper hardware to do both build natively. So I will create a self-hosted runner for WoA build.

bryanpkc:
@kaadam Is this PR ready to merge then? And you will submit another PR for the Windows-on-ARM builds?

kaadam:
> @kaadam Is this PR ready to merge then? And you will submit another PR for the Windows-on-ARM builds?

@bryanpkc I did a tiny modification on this PR, It's ready for review/merge now. Yes, I will upload the Windows on ARM configuration in a follow-up PR.

kaadam:
@bryanpkc I could simplify the x64 and arm64 workflow, so maybe better if I squash two PRs into one.

kaadam:
@bryanpkc I've updated the PR. Could you look at it again please? Currently Windows on ARM config is disabled until an official build will be available. However the workflow is prepared to handle both Windows platform.

bryanpkc (approved):

bryanpkc:
LGTM. I don't think the other maintainers have any objection to this PR.

PR #152, opened 2023-02-28
[Driver] Don't use Classic Flang tools for LLVM IR input

bryanpkc:
LLVM IR input should not be acceptable for Classic Flang tools. This bug was caused by an incorrect merge conflict resolution when upgrading to LLVM 15.

### DIFF
--- a/clang/lib/Driver/Types.cpp
+++ b/clang/lib/Driver/Types.cpp
@@ -175,14 +175,15 @@ bool types::isAcceptedByFlang(ID Id) {
   case TY_PP_F_FreeForm:
   case TY_F_FixedForm:
   case TY_PP_F_FixedForm:
+    return true;
 #else
   case TY_Fortran:
   case TY_PP_Fortran:
-#endif
     return true;
   case TY_LLVM_IR:
   case TY_LLVM_BC:
     return true;
+#endif
   }
 }
 
--- a/clang/test/Driver/flang/llvm-ir-input.f
+++ b/clang/test/Driver/flang/llvm-ir-input.f
@@ -1,0 +1,7 @@
+! Check that LLVM IR input is passed to clang instead of flang1.
+
+! REQUIRES: classic_flang
+! RUN: %clang --driver-mode=flang -S %S/Inputs/llvm-ir-input.ll -### 2>&1 | FileCheck %s
+
+! CHECK-NOT: flang1
+! CHECK: "{{.*}}clang{{.*}}" "-cc1"


pawosm-arm (commented):
This looks good, but I'm curious what use case has made this issue revealed. Can you give some example e.g. in the commit message?

pawosm-arm:
> This looks good, but I'm curious what use case has made this issue revealed. Can you give some example e.g. in the commit message?

Ah, I didn't notice there's a test case

pawosm-arm (approved):

clang/lib/Driver/Types.cpp
```diff
@@ -175,14 +175,15 @@ bool types::isAcceptedByFlang(ID Id) {
   case TY_PP_F_FreeForm:
   case TY_F_FixedForm:
   case TY_PP_F_FixedForm:
+    return true;
 #else
   case TY_Fortran:
   case TY_PP_Fortran:
-#endif
     return true;
```
PeixinQiao, line 182: Should the code in line 182 be removed?

PeixinQiao:
Would some comments in `llvm-ir-input.ll` describing its test purpose be better than one empty file?

clang/lib/Driver/Types.cpp
bryanpkc: No, it shouldn't. Doing that would create an unnecessary diff relative to upstream `release/15.x`.

bryanpkc:
> Would some comments in llvm-ir-input.ll describing its test purpose be better than one empty file?

This file is intentionally left empty to follow LLVM convention. Many input files under `clang/test/Driver/Inputs` are left empty since their contents are irrelevant:
```
$ find clang/test/Driver/Inputs -type f -size 0 \! -name .keep | wc -l
1507
```

PeixinQiao:
> > Would some comments in llvm-ir-input.ll describing its test purpose be better than one empty file?
> 
> This file is intentionally left empty to follow LLVM convention. Many input files under `clang/test/Driver/Inputs` are left empty since their contents are irrelevant:
> 
> ```
> $ find clang/test/Driver/Inputs -type f -size 0 \! -name .keep | wc -l
> 1507
> ```

OK. Thanks. LGTM now.

PeixinQiao (approved):

PeixinQiao:
gentle ping @shivaramaarao

bryanpkc:
@shivaramaarao If you have no objection, I'll merge this PR.

PR #154, opened 2023-03-16
script: update llvm python build script (release_14x)

kaadam:
* Native Ninja is already available for Windows on ARM, so we can use it on both Windows platform.

    Please see: https://pypi.org/project/ninja/#files

* Introduced a new '-i' build option to trigger whether the user wants to install LLVM or not. Previously the script relies on `install_prefix` option which was wrong approach. In this case we still could not defer the install for a later time if install directory was set. With this change the python build script will be consistent with official bash script.

### DIFF
--- a/scripts/build_llvm_project.py
+++ b/scripts/build_llvm_project.py
@@ -26,27 +26,29 @@ def get_arguments():
   buildopt.add_argument('-t', '--target', metavar='ARCH', choices=['X86', 'AArch64', 'PowerPC'], default='X86',
                         help='Control which targets are enabled (%(choices)s) (default: %(default)s)')
   buildopt.add_argument('-p', '--install-prefix', metavar='PATH', nargs='?', default=None, const=False,
-                        help='Install directory (default: do not install)')
+                        help='Specify installation directory (default: platform-specific location)')
+  buildopt.add_argument('-i', '--install', action='store_true', default=False,
+                        help='Install LLVM (default: do not install)')
   buildopt.add_argument('-j', '--jobs', metavar='N', type=int, default=os.cpu_count(),
-                        help='number of parallel build jobs (default: %(default)s)')
+                        help='Number of parallel build jobs (default: %(default)s)')
   buildopt.add_argument('--toolchain', metavar='FILE', default=default_toolchain().as_posix(),
-                        help='specify toolchain file (default: %(default)s)')
+                        help='Specify toolchain file (default: %(default)s)')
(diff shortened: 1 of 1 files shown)


scripts/build_llvm_project.py
```diff
@@ -26,27 +26,29 @@ def get_arguments():
   buildopt.add_argument('-t', '--target', metavar='ARCH', choices=['X86', 'AArch64', 'PowerPC'], default='X86',
                         help='Control which targets are enabled (%(choices)s) (default: %(default)s)')
   buildopt.add_argument('-p', '--install-prefix', metavar='PATH', nargs='?', default=None, const=False,
-                        help='Install directory (default: do not install)')
+                        help='Set install directory (default: platform specific directory)')
+  buildopt.add_argument('-i', '--install', action='store_true', default=False,
+                        help='Do llvm install (default: do not install llvm)')
```
bryanpkc: Please reword these two help messages:
- `Specify installation directory (default: platform-specific location)`
- `Install LLVM  (default: do not install)`

scripts/build_llvm_project.py
```diff
@@ -26,27 +26,29 @@ def get_arguments():
...
+                        help='Clean build')
   buildopt.add_argument('-b', '--build-type', metavar='TYPE', default='Release',
-                        help='set build type (default: %(default)s)')
+                        help='Set build type (default: %(default)s)')
   buildopt.add_argument('-x', '--cmake-param', metavar='OPT', action='append', default=[],
-                        help='add custom argument to CMake')
+                        help='Add custom argument to CMake')
   buildopt.add_argument('-e', '--llvm-enable-projects', metavar='OPT', action='append', default=['clang'] if sys.platform == 'win32' else ['clang', 'openmp'],
-                        help='enable llvm projects to build (in quotation marks separated by semicolons e.g.: "clang;openmp")')
+                        help='Enable llvm projects to build (in quotation marks separated by semicolons e.g.: "clang;openmp")')
```
bryanpkc: Consider make `LLVM` uppercase everywhere.

scripts/build_llvm_project.py
kaadam: Thanks for the review and the suggestion. Updated.

scripts/build_llvm_project.py
kaadam: Done.

bryanpkc (approved):
LGTM

PR #155, opened 2023-03-16
script: update llvm python build script (release_15x)

kaadam:
* Native Ninja is already available for Windows on ARM, so we can use it on both Windows platform.

    Please see: https://pypi.org/project/ninja/#files

* Introduced a new '-i' build option to trigger whether the user wants to install LLVM or not. Previously the script relies on `install_prefix` option which was wrong approach. In this case we still could not defer the install for a later time if install directory was set. With this change the python build script will be consistent with official bash script.


bryanpkc (approved):
LGTM

PR #161, opened 2023-07-13
Add Flang include directory to MSVC toolchain

kaadam:


kaadam:
@bryanpkc @pawosm-arm Hi, Could you take a look at this PR please? Thanks!

clang/lib/Driver/ToolChains/MSVC.cpp
```diff
@@ -689,6 +689,42 @@ void MSVCToolChain::AddSystemIncludeWithSubfolder(
...
+                                  ToolChain::path_list IncludePathList) {
+  std::string ArgValue; // Path argument value
+
+  // Make up argument value consisting of paths separated by colons
+  bool first = true;
+  for (auto P : IncludePathList) {
+    if (first) {
+      first = false;
+    } else {
+      ArgValue += ":";
```
pawosm-arm: using `:` as a separator in the Windows world seems tricky, how would it look like? `"c:\msvc:d:\projects"`, is this approach ok?

kaadam: No, officially the separator is the semicolon ';' on Windows. Thanks for the review. Switching between the terminals, (mingw, ms command prompt) confuses me sometimes. But ':' is also works. I will update the separator.

kaadam:
@bryanpkc, @pawosm-arm I've updated the PR, could you look at it again, please?

pawosm-arm (approved):

bryanpkc (approved):

PR #165, opened 2023-09-15
[Driver] Fix flang driver preprocessor issue

bryanpkc:
The `-save-temps` option has been causing the driver to run `flang1`/`flang2` on the input file twice, and the `fortran-preprocessor.f90` test was unable to detect the error, because it had used `-c` instead of `-x` in a couple of `RUN` commands. This patch fixes the test and the driver to make sure that the frontend is not invoked multiple times with `-save-temps`.

### DIFF
--- a/clang/test/Driver/fortran-preprocessor.f90
+++ b/clang/test/Driver/fortran-preprocessor.f90
@@ -2,7 +2,7 @@
 
 ! -cpp should preprocess as it goes, regardless of input file extension
 ! RUN: %flang -cpp -c -DHELLO="hello all" -### %s 2>&1 | FileCheck %s --check-prefixes=ALL,CPP,PP
-! RUN: %flang -cpp -c -DHELLO="hello all" -### -c f95-cpp-input %s 2>&1 | FileCheck %s --check-prefixes=ALL,CPP,PP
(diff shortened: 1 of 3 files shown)


clang/test/Driver/fortran-preprocessor.f90
```diff
@@ -2,7 +2,7 @@
 
 ! -cpp should preprocess as it goes, regardless of input file extension
 ! RUN: %flang -cpp -c -DHELLO="hello all" -### %s 2>&1 | FileCheck %s --check-prefixes=ALL,CPP,PP
-! RUN: %flang -cpp -c -DHELLO="hello all" -### -c f95-cpp-input %s 2>&1 | FileCheck %s --check-prefixes=ALL,CPP,PP
+! RUN: %flang -cpp -c -DHELLO="hello all" -### -x f95-cpp-input %s 2>&1 | FileCheck %s --check-prefixes=ALL,CPP,PP
```
pawosm-arm, line 5: whoa, how it was supposed to work?!
I mean it's good it's finally sorted.

pawosm-arm (approved):

clang/test/Driver/fortran-preprocessor.f90
bryanpkc: If you removed the `-###` option, you would see that the command actually failed since `f95-cpp-input` is not an existing file.

PR #168, opened 2023-12-21
[workflows] Update GitHub actions to use modern tools

bryanpkc:
Build classic-flang-llvm-project and flang with GCC 12 and Clang 14/15 on Ubuntu 22.04.


bryanpkc:
Note to reviewers: The AArch64 test jobs won't run successfully on this PR since they depend on the Docker container which will only be available after this PR is merged, and the new container image is generated.

pawosm-arm (approved):

PR #174, opened 2024-01-24
[workflows] Fix typo in pre-compile_llvm.yml; NFC

bryanpkc:


pawosm-arm (approved):
nasty typo! a single approval should suffice in such case.

PR #180, opened 2024-04-21
Fix for issue 1440 (make Mbackslash as the default option in flang)

shivaramaarao:
Through this commit Mbackslash will be the default option. By default, escape characters are treated as regular characters unless Mnobackslash is specified.

### DIFF
--- a/clang/lib/Driver/ToolChains/ClassicFlang.cpp
+++ b/clang/lib/Driver/ToolChains/ClassicFlang.cpp
@@ -240,18 +240,31 @@ void ClassicFlang::ConstructJob(Compilation &C, const JobAction &JA,
     }
   }
 
-  // Treat backslashes as regular characters
-  for (auto Arg : Args.filtered(options::OPT_fno_backslash, options::OPT_Mbackslash)) {
-    Arg->claim();
+  if (auto *A =
+          Args.getLastArg(options::OPT_fno_backslash, options::OPT_Mbackslash,
+                          options::OPT_fbackslash, options::OPT_Mnobackslash)) {
+    for (auto Arg :
+         Args.filtered(options::OPT_fno_backslash, options::OPT_Mbackslash,
+                       options::OPT_fbackslash, options::OPT_Mnobackslash)) {
+      Arg->claim();
+    }
+    // Treat backslashes as regular characters
+    if (A->getOption().matches(options::OPT_fno_backslash) ||
+        A->getOption().matches(options::OPT_Mbackslash)) {
+      CommonCmdArgs.push_back("-x");
+      CommonCmdArgs.push_back("124");
+      CommonCmdArgs.push_back("0x40");
+    }
+    // Treat backslashes as C-style escape characters
+    if (A->getOption().matches(options::OPT_fbackslash) ||
+        A->getOption().matches(options::OPT_Mnobackslash)) {
+      CommonCmdArgs.push_back("-y");
+      CommonCmdArgs.push_back("124");
+      CommonCmdArgs.push_back("0x40");
+    }
+  } else {
+    // By default treat backslashes as regular characters
     CommonCmdArgs.push_back("-x");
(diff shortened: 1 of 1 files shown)


clang/lib/Driver/ToolChains/ClassicFlang.cpp
```diff
@@ -250,11 +252,19 @@ void ClassicFlang::ConstructJob(Compilation &C, const JobAction &JA,
 
   // Treat backslashes as C-style escape characters
   for (auto Arg : Args.filtered(options::OPT_fbackslash, options::OPT_Mnobackslash)) {
+    BackSlashOptionPresent = true;
```
bryanpkc: Shouldn't these two `for` loops be combined, to handle all four options at once? Currently when the `-fbackslash -fno-backslash` command-line options are specified in that order, the driver gives this surprising behaviour (overriding `-x 124 0x40` with `-y 124 0x40`):
```
".../bin/flang1" foo.f90 -opt 0 -terse 1 -inform warn -nohpf -nostatic -cmdline "'+flang -fbackslash -fno-backslash foo.f90 -v'" -inform warn -x 124 0x40 -y 124 0x40 -y 129 2 -ieee 1 ...
```

bryanpkc (changes requested):
The option processing is buggy and should be fixed. The test case `clang/test/Driver/flang/classic-flang.f95` should be updated to cover the relevant options.

clang/lib/Driver/ToolChains/ClassicFlang.cpp
```diff
@@ -240,20 +240,39 @@ void ClassicFlang::ConstructJob(Compilation &C, const JobAction &JA,
...
+  if (auto *A = Args.getLastArg(options::OPT_fno_backslash, 
+                                options::OPT_Mbackslash,
+                                options::OPT_fbackslash, 
+                                options::OPT_Mnobackslash)) {
+    for (auto Arg : Args.filtered(options::OPT_fno_backslash, 
+                                  options::OPT_Mbackslash)) {
+      Arg->claim();
+    }
+    for (auto Arg : Args.filtered(options::OPT_fbackslash, 
+                                  options::OPT_Mnobackslash)) {
```
bryanpkc: Why does this code claim the options in two separate loops? Can't it claim all of them in a single loop?

clang/lib/Driver/ToolChains/ClassicFlang.cpp
```diff
@@ -240,20 +240,39 @@ void ClassicFlang::ConstructJob(Compilation &C, const JobAction &JA,
...
+      A->claim();
+      CommonCmdArgs.push_back("-y");
+      CommonCmdArgs.push_back("124");
+      CommonCmdArgs.push_back("0x40");
+    }
+  } else {
+    // By default treat backslashes as regular characters
+      CommonCmdArgs.push_back("-x");
+      CommonCmdArgs.push_back("124");
+      CommonCmdArgs.push_back("0x40");
```
bryanpkc: Incorrect indentation. Please use `clang-format` on the diff.

clang/lib/Driver/ToolChains/ClassicFlang.cpp
```diff
@@ -240,20 +240,39 @@ void ClassicFlang::ConstructJob(Compilation &C, const JobAction &JA,
...
+      Arg->claim();
+    }
+    for (auto Arg : Args.filtered(options::OPT_fbackslash, 
+                                  options::OPT_Mnobackslash)) {
+      Arg->claim();
+    }
+    // Treat backslashes as regular characters
+    if (A->getOption().matches(options::OPT_fno_backslash) ||
+        A->getOption().matches(options::OPT_Mbackslash)) {
+       A->claim();
```
bryanpkc: `A` has already been claimed by this point.

bryanpkc (approved):
LGTM.

pawosm-arm (approved):

PR #179, opened 2024-04-21
Fix for Issue #1413

shivaramaarao:
PGMATH has AVX512 runtime functions and can be executed only when the application is compiled in avx512 mode. The VecFuncs.def has no information about the TargetOptions and avx512 functions are selected even in avx2 mode. This issue is fixed by creating separate table for AVX512 functions and using them only when avx512 mode is specified.

PR #184, opened 2024-07-26
Pgmath avx512 fix

shivaramaarao:
This is cherry-pick of  #179  from release_17x branch.

PGMATH has AVX512 runtime functions and can be executed only when the application is compiled
in avx512 mode. The VecFuncs.def has no information about the TargetOptions and avx512 functions
are selected even in avx2 mode. This issue is fixed by creating separate table for AVX512 functions
and using them only when avx512 mode is specified.


shivaramaarao:
some issue with pull request.

shivaramaarao:
some issue in pull request.

PR #187, opened 2024-07-26
Fix for issue #1440 (cherry-pick #180 to release_18x)

shivaramaarao:
Through this commit Mbackslash will be the default option. By default, escape characters are treated as regular characters unless Mnobackslash is specified.

There is also the issue of treating both Mbackslash and Mnobackslash options independently. This is addressed in this patch and the testcases in flang repository will test this feature.


bryanpkc (approved):
LGTM; identical to #180.

pawosm-arm (approved):

PR #218, opened 2026-07-10
[workflows] Unify X86 and ARM64 builds; other minor improvements

bryanpkc:
Unifiy the X86 and ARM64 builds into one reusable workflow that runs on Ubuntu 24.04 runners, to match upstream llvm-project (`release/21.x`).

Update the bootstrapping compilers to Clang 17/18 and GCC 13.

Stop using CCache when building Flang, because the Clang/LLVM binaries have changed.

Test Flang with the long test suite, to prevent introducing LLVM bugs that may break subsequent Flang PR checks.


bryanpkc:
The CI jobs are failing due to flang-compiler/flang#1453, which is a known intermittent issue. I'm going to merge this PR and fix the test case in a separate PR.

